PR #7, opened 2024-06-04
Added basic classes for handling config file parsing

QCHR1581:
Added following classes:

- ConfigFile -> contains vector of ServerConfig objects
- ServerConfig -> contains needed parameters for a server (incl. a vector of Location objects)
- Location -> contains needed parameters for a location


inc/Location.hpp
```diff
@@ -0,0 +1,24 @@
...
+#include <vector>
+
+class Location {
+public:
+    Location();
+    ~Location();
+
+private:
+    std::string m_path;
+    std::string m_root;
```
gwolf-011235, line 15: I would suggest to init with a default value, where there is one in nginx (or even always?)

For example: `std::string m_root = "html";`
I think this can prevent some headaches while parsing, because you don't need to explicitly set them to default, since they are set anyway

inc/Location.hpp
```diff
@@ -0,0 +1,24 @@
...
+    std::string m_index;
+    std::string m_cgi_ext;
+    std::string m_cgi_path;
+    bool m_autoindex;
+    std::vector<std::string> m_allowed_methods;
```
gwolf-011235, line 20: I don't think we need to store the allowed methods as strings. 
Since there are only a small set of methods we could define them in an enum and then use a bool array or bit field or whatever to simply check.

For example (not tested :D )

> define enum

`enum METHOD { METHOD_GET, METHOD_POST, METHOD_DELETE, METHOD_COUNT };`

> in the class we can use count for our array

`bool m_allowed_methods[METHOD_COUNT];`

> then in code check if the method is allowed

`if (m_allowed_methods[METHOD_GET]) { do stuff };`

inc/ServerConfig.hpp
```diff
@@ -0,0 +1,25 @@
...
+class ServerConfig {
+public:
+    ServerConfig();
+    ~ServerConfig();
+
+private:
+    std::string m_server_name;
+    std::string m_host;
+    std::string m_index;
+    std::string m_root;
```
gwolf-011235, line 18: I think we should save every directive only in the lowest context. For root this would be "location". 

You can set root in server and in location context (also html). 
Since you don't need to set a location context at all, I think you would then save root in here. But this means that we have to check for every location if there is a root, and if not use the one from the server block. We also need to check if there is a location context and if not fall back.

In my mind:
- we set a standard location, if a server level root directive is declared. 
- In there we save the root directive, which is declared in server context
- If we have one (or more) location context defined, we set the root as declared in location context
- if not declared we set it as declared in server context
- if not declared we set it as default value. 

This way when we check which location should be used:
- if no location was defined we can still use our default location
- if one was declared, we check if it fits
- if root was also defined on server level, we create additional default location, which means we have two locations

I think the same stuff could also apply to other directives, but I don't see another one right now.

inc/Location.hpp
```diff
@@ -0,0 +1,24 @@
...
+    std::map<unsigned short, std::string> m_return;
+};
+
+#endif
```
gwolf-011235, line 24: Newline :D

inc/ServerConfig.hpp
```diff
@@ -0,0 +1,25 @@
...
+    unsigned short m_port;
+    unsigned long m_max_body_size;
+    std::map<unsigned short, std::string> m_error_page;
+    std::vector<Location> m_locations;
+};
+
+#endif
```
gwolf-011235, line 25: Newline :D

gwolf-011235 (changes requested):
Good structure layout - some improvements

One last time I'll bring it up: I still think that the config file could just be a struct (of structs). It only holds data, which won't be changed during runtime (I think?). So the values should mostly be const and not changeable anyway (no need for setters).

The ConfigFileParser() could be a class which receives a path to config file, parses it and writes it to a ConfigFile struct. This struct is just a data blob of (unchangeable) values which can be passed around and referenced and read from.

CPP FAQ also points in that direction: [Difference class and struct](https://isocpp.org/wiki/faq/classes-and-objects#struct-vs-class)

inc/ConfigFile.hpp
```diff
@@ -0,0 +1,35 @@
...
+    const std::string m_index;
+    const std::string m_cgiExt;
+    const std::string m_cgiPath;
+    const bool m_autoindex;
+    bool m_allowed_methods[MethodCount];
+    std::map<unsigned short, std::string> m_return;
+};
+
+struct ServerConfig {
+    const std::string m_serverName;
```
u413-284-si, line 25: see above comment on const type and variable prefix

inc/ConfigFileParser.hpp
```diff
@@ -0,0 +1,12 @@
+#pragma once
+
+#include "ConfigFile.hpp"
+
+class ConfigFileParser {
+public:
+    ConfigFileParser(std::string config_file);
```
u413-284-si, line 7: following Webkit style: std::string configFile

inc/ConfigFile.hpp
```diff
@@ -0,0 +1,35 @@
...
+    const std::string m_host;
+    const unsigned short m_port;
+    const unsigned long m_maxBodySize;
+    std::map<unsigned short, std::string> m_errorPage;
+    std::vector<Location> m_locations;
+};
+
+struct ConfigFile {
+    std::vector<ServerConfig> m_serverConfigs;
```
u413-284-si, line 34: prefix

u413-284-si (changes requested):
Like the structure with a separate ConfigFile header. The separation of the ConfigFile struct from the Parser class becomes more apparent this way.

Albeit most comments focus on formatting/style conventions, there are two major functional points to be considered:

1. Usage of const declaration of variables in structs: When variables in a struct are declared as const, it means that the values of those variables cannot be modified, only when they are initialized. This would be not practical to implement in the parser as you would probably want to read the config file string piece by piece and assign the variables on the go.
2. Usage of bool array: As explained in detail in the remarks, the advantage of such a tool is doubtful in our use case. I would rather see a switch case in combination with the enum where we execute certain cmds according to the appropriate case.

 Method currentMethod = GET;
 
 switch (currentMethod) {
 case GET: // do sth
 case POST: // do sth else
 ...
 }

inc/ConfigFile.hpp
```diff
@@ -0,0 +1,35 @@
...
+    MethodCount };
+
+struct Location {
+    const std::string m_path;
+    const std::string m_root;
+    const std::string m_index;
+    const std::string m_cgiExt;
+    const std::string m_cgiPath;
+    const bool m_autoindex;
+    bool m_allowed_methods[MethodCount];
```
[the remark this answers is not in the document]
gwolf-011235: I understood this directive as methods, which are allowed. We have to implement three methods, but this doesn't mean we have to allow them all. The bool array would track which methods are allowed to be used.

I didn't find a nginx directive called allowed_methods, but I think the corresponding is [limit_except](https://nginx.org/en/docs/http/ngx_http_core_module.html#limit_except). However this is in location context, which makes sense to me: depending on the location you can do different things.

inc/ConfigFile.hpp
```diff
@@ -0,0 +1,35 @@
+#pragma once
+
+#include <iostream>
+#include <map>
+#include <vector>
+
+enum Method { MethodGet,
```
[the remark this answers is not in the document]
gwolf-011235: StyleGuide disagrees :o [names enum members](https://webkit.org/code-style-guidelines/#names-enum-members)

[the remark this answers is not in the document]
u413-284-si: Oha, missed that line. Then leave it.

inc/ConfigFile.hpp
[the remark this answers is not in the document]
u413-284-si: Where does this directive then come from? If it is not part of nginx.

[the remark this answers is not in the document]
QCHR1581: This is not a directive specified in the doc nginx itself. But I found [this module](https://github.com/pschultz/ngx_http_allow_methods_module). 

It made in my opinion sense if there exists a module for this specific purpose (define which methods are allowed) to chose this for inspiration.

But I think that limit_except suits better for this (I didn't find this directive at first)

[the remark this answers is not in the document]
u413-284-si: Ok, so depending on the location, we restrict the access of the three methods (GET, POST, DELETE) that we implement with this limit_except directive. Makes sense.

u413-284-si (approved):
Looks good now! There are still a few indentations which are not aligned, but this can be fixed easily with Gabe's neat formatter later on ;)

gwolf-011235 (approved):

PR #36, opened 2024-08-03
feat: log request parsing with Logger

u413-284-si:
Use Logger macros to log major parsing steps and thrown exceptions for the RequestParser.

Deliberately no closing keywords to keep correlated issue open to be able to create a similar branch for the ConfigFileParser.

### DIFF
--- a/src/RequestParser.cpp
+++ b/src/RequestParser.cpp
@@ -1,4 +1,5 @@
 #include "RequestParser.hpp"
+#include "Log.hpp"
 #include <cstddef>
 
 /* ====== HELPER FUNCTIONS ====== */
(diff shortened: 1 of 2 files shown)


src/RequestParser.cpp
```diff
@@ -282,6 +283,10 @@ void RequestParser::parseHttpRequest(const std::string& requestString, HTTPReque
 	requestLine = parseVersion(requestLine, request);
 	checkForCRLF(requestLine, request);
 
+    LOG_DEBUG << "Parsed method: " << request.method;
```
gwolf-011235, line 286: Since method is an enum, we could make an ostream overload to print out the string for the method

gwolf-011235 (approved):
Looks good!

src/RequestParser.cpp
u413-284-si: Added insertion overload for enum method.

gwolf-011235: LogData doesn't seem like the right place, since it already has a templated ostream receiver: if there is an ostream overload, LogData can use it.

In #39 I implemented it like the existing ostream insertion operators for Config file and renamed the file LogConfigFile.cpp to reflect that it contains different ostream inserters

u413-284-si: You mean #38, not #39, right? 

Alright, if you created a general file for all ostream overloads I will remove the last commit and merge directly. The new file will be included via the LogData.hpp after your PR is merged.

u413-284-si: Need this conversation resolved for merge.

gwolf-011235: Yes you're right: #38

Issue #22, opened 2024-07-17
Implement Connection class


gwolf-011235:
The virtual servers and clients can be abstracted in a generic Connection class. This can be passed to the void*ptr of the data union of the epoll event. 

The advantage is to provide more context to server or client events.

The Connection will also be equipped with an event handler to dispatch to the appropriate event, e.g. accept() for server connections and read or write for client.

PR #40, opened 2024-08-04
14 core call epoll with timeout value

### LINKED ISSUE #14: core: call epoll with TIMEOUT value
Calling epoll with a timeout prevents the server from waiting endlessly for a connection which may be broken or has faulty requests, such as a chunked body which is missing the zero-size chunk at the end. The server would be stuck in an endless loop otherwise, expecting to find this zero chunk to indicate the end of the body. Having a timeout will help to break out of the loop and take appropriate actions e.g. closing the connection.

EDIT:
Alternatively implement a check in the parseChunkedBody() loop for current body length vs. max. allowed body length (defined by config file or default value). Break out of loop if check condition reached.

u413-284-si:
Add timeout value to epoll_wait to activate epoll when there have been no events for said time span.
Add tracking of last_active point in time for each connection. Epoll checks for each connection whether a predetermined idle time threshold has been breached. In case yes, the connection is closed.

This prevents idle connections to stay open endlessly.

Closes #14

### DIFF
--- a/inc/Server.hpp
+++ b/inc/Server.hpp
@@ -23,7 +23,6 @@
 
 /* ====== DEFINITIONS ====== */
 
-#define MAX_EVENTS 10
 #define PORT 8080
 #define BUFFER_SIZE 1024
 #define CONNECTION_QUEUE 10
@@ -35,6 +34,7 @@ private:
 	int m_serverSock;
 	int m_epfd;
 	std::map<int, std::string> m_requestStrings;
+    std::map<int, time_t> m_connectionLastActive;
 
 	void acceptConnection() const;
 	void handleConnections(int clientSock, RequestParser& parser);
--- a/inc/Server.hpp
+++ b/inc/Server.hpp
@@ -34,7 +34,7 @@ private:
 	int m_serverSock;
 	int m_epfd;
 	std::map<int, std::string> m_requestStrings;
-    std::map<int, time_t> m_connectionLastActive;
+	std::map<int, time_t> m_connectionLastActive;
 
 	void acceptConnection() const;
 	void handleConnections(int clientSock, RequestParser& parser);
--- a/src/Server.cpp
+++ b/src/Server.cpp
@@ -1,5 +1,6 @@
 #include "Server.hpp"
 #include "ConfigFile.hpp"
+#include "Log.hpp"
 
 /* ====== CONSTRUCTOR/DESTRUCTOR ====== */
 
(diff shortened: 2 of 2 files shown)


inc/Server.hpp
```diff
@@ -35,6 +34,7 @@ class Server {
 	int m_serverSock;
 	int m_epfd;
 	std::map<int, std::string> m_requestStrings;
+	std::map<int, time_t> m_connectionLastActive;
```
gwolf-011235, line 37: Since there are several datapoints referring to a specific connection (connected server, time, bytesReceived) I would manage it in its own class like done in #22
I didn't move the buffer String map (yet in #22 ), but would also opt to do so

src/Server.cpp
```diff
@@ -122,13 +123,29 @@ Server& Server::operator=(const Server& ref)
 void Server::run()
 {
 	RequestParser parser;
+	const int maxEvents = 10;
+	const int timeout = 5000; // epoll timeout in milliseconds
+	const int acceptedIdleTime = 10; // time in seconds until an idle connection is closed
```
gwolf-011235, line 128: In #22 I implemented member variables and static const ints for default values. I think this makes it more flexible.
I dont think we would make maxEvents or timeout configurable but I think acceptedIdleTime is a valid configure point

src/Server.cpp
```diff
@@ -122,13 +123,29 @@ Server& Server::operator=(const Server& ref)
...
 
 	while (true) {
-		struct epoll_event events[MAX_EVENTS];
+		struct epoll_event events[maxEvents];
```
gwolf-011235, line 131: I know this is only changed because no Macro but funny thing:
In #22 I implemented this as std::vector to avoid C-style array. In newer standard one would just use std::array, but this way there is no array subscript access :D

src/Server.cpp
```diff
@@ -122,13 +123,29 @@ Server& Server::operator=(const Server& ref)
...
 		// Blocking call to epoll_wait
-		int nfds = epoll_wait(m_epfd, events, MAX_EVENTS, -1);
+		int nfds = epoll_wait(m_epfd, events, maxEvents, timeout);
 		if (nfds < 0)
 			throw std::runtime_error("epoll_wait");
+		if (nfds == 0) {
+			LOG_WARN << "Timeout: no events received within " << timeout << " ms.";
```
gwolf-011235, line 137: In #22 I have it as LOG_DEBUG.
I didn't see it as warning since not getting any (new) events seems "normal" to me for a webserver. We can do stuff like handleTimeout when no event occured.

src/Server.cpp
```diff
@@ -122,13 +123,29 @@ Server& Server::operator=(const Server& ref)
...
+
+			time_t currentTime = std::time(NULL);
+			for (std::map<int, time_t>::iterator iter = m_connectionLastActive.begin();
+				 iter != m_connectionLastActive.end(); iter++) {
+				if (currentTime - iter->second > acceptedIdleTime) {
+					close(iter->first);
+					m_connectionLastActive.erase(iter);
+				}
+			}
+			continue;
```
gwolf-011235, line 147: In #22 I do this check every time epoll_wait() returns, not only if no event is received. If we return from it we might as well check for timeout.

gwolf-011235 (changes requested):
Large overlap with #22. I suggest to handle the PR of #22 = #37 beforehand and add comments for different implementation needs

src/Server.cpp
u413-284-si: Legit.

src/Server.cpp
u413-284-si: You have it as LOG_INFO even. But is fine by me.

src/Server.cpp
u413-284-si: Even better!

src/Server.cpp
u413-284-si: I guess it's a more centralised place to store the variables.

inc/Server.hpp
u413-284-si: WDYM move the buffer string map?

u413-284-si:
As most of this is implemented in branch #37, I guess we can drop this PR.

inc/Server.hpp
gwolf-011235: Move the buffer in the Connection class. Like a Connection class has a member buffer

src/Server.cpp
gwolf-011235: Leave as LOG_INFO or switch LOG_DEBUG? Right now LOG_DEBUG makes more sense to me :D

inc/Server.hpp
u413-284-si: Yeah makes sense.

u413-284-si:
> How do we drop a PR? We never did that

I would just close the PR w/o merging and then delete the branch.

PR #41, opened 2024-08-04
38 create ostream overload for enum method

gwolf-011235:
- Rename LogConfigFile.cpp to LogOstreamInserters.cpp > to store all stream inserters in there.
- Ostream inserter definitions reside in the header file where their respective enum/struct is defined. This is to reduce dependencies.
- The file LogOstreamInserters.cpp includes these defintions from the headers from enum/struct
- New inserters for 
  - enum Method
  - enum StatusCode
  - struct URI
  - struct HTTPRequest
  - struct HTTPResponse
- Added tests


src/LogOstreamInserters.cpp
```diff
@@ -0,0 +1,200 @@
...
+		break;
+	case MethodPost:
+		ostream << "POST";
+		break;
+	case MethodDelete:
+		ostream << "DELETE";
+		break;
+	case MethodCount:
+		ostream << "enum MethodCount";
+		break;
```
u413-284-si, line 101: What about the default case, when none of the above cases are hit? We should include

```
default:
       ostream << "UNKNOWN METHOD";
```

src/LogOstreamInserters.cpp
```diff
@@ -0,0 +1,200 @@
...
+	case StatusInternalServerError:
+		ostream << "500 Internal Server Error";
+		break;
+	case StatusMethodNotImplemented:
+		ostream << "501 Method Not Implemented";
+		break;
+	case StatusNonSupportedVersion:
+		ostream << "505 HTTP Version Not Supported";
+		break;
+	}
```
u413-284-si, line 143: Same as above, would provide a default case.

test/test_OstreamInserters.cpp
```diff
@@ -0,0 +1,223 @@
+#include <gtest/gtest.h>
+
+#include "ConfigFile.hpp"
+#include "LogOstreamInserters.hpp"
```
u413-284-si, line 4: Header file is missing. Leads to failed test compilation.

u413-284-si (changes requested):
A header file was somehow missing, maybe deleted by accident?
Besides that, a general point regarding use of switch cases.

Otherwise nice work!

src/LogOstreamInserters.cpp
gwolf-011235: I deliberately did not include a default: clang-tidy fires a warning if not all cases of an enum over a switch are covered. It gets deactivated when there is a default case, since default catches everything else. 

I researched a bit and found this interesting thread:
--> https://stackoverflow.com/questions/2201493/using-default-in-a-switch-statement-when-switching-over-an-enum

The highest voter would throw an exception (I think as default case), others say assert() and always have default.

LLVM says [Don’t use default labels in fully covered switches over enumerations](https://llvm.org/docs/CodingStandards.html#id48)
They mention -Wswitch which will print a warning when not all cases of an enum are covered in a switch.

Another argument is to throw/abort if none of the cases are reached, not as default but after the switch ran through. Argument being that memory corruption could happen, and if it does aborting is the only option since the program behaves abnormal.

Googles C++ Style Guide explicitly makes an exception for switch cases over enums: [If not conditional on an enumerated value, switch statements should always have a default case (in the case of an enumerated value, the compiler will warn you if any values are not handled).](https://google.github.io/styleguide/cppguide.html#Switch_Statements)

I would prefer to throw outside of the switch case and add -Wswitch to our Makefile
Then everytime we switch over an enum we never add an default, if a new case gets added we get an compile error

test/test_OstreamInserters.cpp
gwolf-011235: Yes. This was after the last refactor, where I deleted the "centralised" header. 

Added missing headers, now test compiles and runs through

src/LogOstreamInserters.cpp
gwolf-011235: See above

src/LogOstreamInserters.cpp
u413-284-si: I can get behind leaving the default case away in order to get the warnings whenever an enum is added and not handled in the switch.

But then we need an additional var which is not set when none of the cases are hit, and will then throw an exception after the switch. That is how I understand the intention of the commenter referring to LLVM and Google.

src/LogOstreamInserters.cpp
u413-284-si: Can be treated as above.

src/LogOstreamInserters.cpp
gwolf-011235: I think LLVM and Google are not that strict, they just say let the compiler warn you.

The stricter variant comes from a commenter of the SO thread who says "what if your program has memory issues and a wrong value gets passed?

We could add an `assert(value >= minEnum && value <= maxEnum)` 
But I think this reaches a level of caution which we don't follow anywhere else

u413-284-si: I would integrate either the assert() or throw an exception.  Our goal is to write clean and safe code, and here we have identified a learning/practice opportunity. That is also the root reason why we follow (most) of the clang tidy checks and put in the effort to make the changes.

gwolf-011235: Added assert() before switch cases on Method and StatusCode

u413-284-si (approved):

QCHR1581 (approved):
No issues found. Good work!

PR #47, opened 2024-08-26
Update dev container

gwolf-011235:
### Summary
1. enhances the dev container with new binaries
2. add new Make targets test_sani, coverage, doxycheck

####  1. Dev container
- Use latest base image of debian-12: mcr.microsoft.com/devcontainers/cpp:debian-12 > beforehand it was fixed at version 1
- Link c++ to clang++ > beforehand it was g++
- Added new tools: 
  - **netcat-openbsd**: a versatile command-line tool used for reading from and writing to network connections; openbsd version has additional functionality, like support for IPv6
  - **telnet**: also a command-line utility for remotely accessing and managing network devices over TCP / alternative to netcat
  - **siege**: command-line tool for load testing and benchmarking web servers by simulating multiple users accessing a web application simultaneously
  - **kcov**:  a code coverage tool for Linux that collects and visualizes coverage data for compiled programs by running them and tracking which lines of code are executed.
- Build **valgrind** from source to use newest version 3.23.0, instead of 3.19 shipped with debian > supports gdwarf-5
- add extension **Coverage Gutters** to devcontainer and add some basic settings for it

#### 2. Makefile
- Add conditionals for variables OBJ_DIR and OBJS to be able to compile with different flags and not need to recompile everytime. If target NAME is built, creates a subdirectory. If target test is built, creates it's own subdirectory
- new target **test_sani**: compiles TEST additionally with -fsanitize=address,undefined. Produces unittest_sani
- new target **doxycheck**: compiles NAME with -Wdocumentation which warns if doxygen comments are not conforming to return type or function params
- new target **coverage**: Creates coverage report readable by extension Coverage Gutters in ".vscode/coverage"
- updated target **comp**: adds compilation of test to compile_commands


QCHR1581 (changes requested):
Great idea to update the dev container!

I like that you also included additional helpful tools. I have tested nginx several times using telnet and our web server for testing purposes. Additionally, kcov is another useful tool that we discussed in the online meeting, and the extension for it is also a nice addition.

Furthermore, the new targets make certain functionalities easier to use.

Regarding kcov, I think it would be cool to also generate a report directory that contains all the necessary files to view the report using the Live Preview extension. This would make it easier for us to have a comprehensive overview of the coverage report.

It is not necessary to implement that but I think it would be a nice bonus.

.devcontainer/devcontainer.json
```diff
@@ -10,8 +10,22 @@
 			"extensions": [
 				"ms-vscode.cpptools-extension-pack",
 				"llvm-vs-code-extensions.vscode-clangd",
-				"cschlosser.doxdocgen"
-			]
+				"cschlosser.doxdocgen",
+				"ryanluker.vscode-coverage-gutters"
```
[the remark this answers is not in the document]
gwolf-011235: Interestingly enough the extension Coverage Gutter depends on this extension, as it uses it for its preview functionality.

Makefile
```diff
@@ -210,6 +239,16 @@ check_bear_installed:
 		echo >&2 "bear is not installed. Please install bear to continue."; exit 1; \
 	}
 
+# Create coverage report to display with coverage gutter
+.PHONY: coverage
+coverage: | $(COV_DIR)
+	@printf "$(YELLOW)$(BOLD)Creating coverage report$(RESET) [$(BLUE)$@$(RESET)]\n"
+	$(SILENT)kcov --cobertura-only $(COV_DIR) ./$(TEST)
```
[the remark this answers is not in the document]
gwolf-011235: I added it on an extra line, and put the coverage report in the same COV_DIR. 

We could host coverage data on something like https://coveralls.io/ but I think this would be better to do per Github action. So I don't think there is a disadvantage having cov.xml and index.html in the same COV_DIR 🤔 

Coverage Gutter has a Live Preview functionality. The search for coverage data is limited to COV_DIR by design: this should make it a bit faster, since the extension doesn't have to travel the whole project.
Now the target creates cov.xml and also the normal reports with index.html.
You can launch a preview it with Ctrl+Shift+6

Also excluded the system headers and our testfiles - which made the coverage drop significantly :o 

Also added COV_DIR to fclean.

.devcontainer/devcontainer.json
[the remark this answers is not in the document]
QCHR1581: You are right. 

I thought that I installed that extension myself in the dev container, because I used it some time before. 
But it gets also installed, when only the Coverage Gutter extensions gets specified in the devcontainer.json file.

Makefile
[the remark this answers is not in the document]
QCHR1581: Great, aside from the coverage drop :(

I don't see any issue with putting everything in the coverage dir. My initial idea was to create a separate dir for organizational purposes, but it's not strictly necessary.

.devcontainer/Dockerfile
```diff
@@ -1,30 +1,68 @@
...
+		libiberty-dev \
+	# Remove valgrind because we build newer version
+	&& apt-get remove -y \
+		valgrind \
+	&& rm -rf /var/lib/apt/lists/*
 
-# Setup googletest
-RUN mkdir -p $HOME/build \
-	&& cd $HOME/build \
+# Build googletest, kcov, and valgrind
```
u413-284-si, line 42: Would separate these cmds for 
- googletest
- kcov
- valgrind
Provides clearer structure and also permits execution of individual packages in case a previous cmd fails for some reason.

Makefile
```diff
@@ -53,25 +74,22 @@ DEP_DIR := $(OBJ_DIR)/dep
...
 
+# Directory for coverage report
+COV_DIR := .vscode/coverage
+
 # ******************************
 # *     Vars for compiling     *
 # ******************************
 
-CXX := c++
+CXX := clang++
```
u413-284-si, line 84: Do we know that the 42 pcs are using clang as compilers?

Makefile
```diff
@@ -141,33 +165,37 @@ LOG_PERF = $(LOG_FILE)_perf.data
...
-# Set file counter to number of test files + object files
-$(TEST): TOTAL_FILES := $(words $(TEST_OBJS))
-$(TEST): $(TEST_OBJS)
-	@printf "$(YELLOW)$(BOLD)link $(TEST)$(RESET) [$(BLUE)$@$(RESET)]\n"
-	$(SILENT)$(CXX) $(TEST_OBJS) -lgtest -lgmock -lgtest_main -o $(TEST)
-	@printf "$(YELLOW)$(BOLD)compilation successful$(RESET) [$(BLUE)$@$(RESET)]\n"
-	@printf "$(BOLD)$(GREEN)$(TEST) created!$(RESET)\n"
+# Alias for creating unittests with sanitizers enabled
+.PHONY: test_sani
+# Reconfigure flags for linking with gtest and sanitizers
```
u413-284-si, line 194: So we are checking the unit tests for memory leaks?

u413-284-si (changes requested):
Nice addition of various functionalities!

Some open points/questions regarding compilation choices and checks.

.devcontainer/Dockerfile
gwolf-011235: Did you mean separate into three RUN commands?
The reason behind combining was to reduce Docker layers: I read that you should reduce layers to keep the image size small. 

But it makes no difference here: image size stayed at 2.48gb, independent if one or three RUN commands. 
So splitting makes sense since we can cash the different steps. 
Made three RUN commands

I moved the valgrind building up, since it takes the longest to build.

I also removed apt install for gtest/gmock and use the same style of cmake, make, make install > reduced image size to 2.47gb

>  in case a previous cmd fails

If you meant separate the three commands inside the RUN command: I don't think we want to continue if something fails. The we would have a half init container

Makefile
gwolf-011235: Tested it right now: yes.

But this was an oversight: I had it in there to check if it makes a difference. Reverted back to c++. Also subject states we have to use c++

However: we should still try to compile on both clang and gcc. Maybe the link gets changed or something?

Makefile
gwolf-011235: This target compiles the unittests with Address sanitizer (ASan) and Unbdefined Behavior Sanitizer (UBSan)
- [ASan ](https://github.com/google/sanitizers/wiki/AddressSanitizer) can detect some memory leaks but its better at detecting memory errors, like use after free, stack buffer overflow etc
- [UBSan](https://clang.llvm.org/docs/UndefinedBehaviorSanitizer.html) detects undefined behavior

The sanitizers are more or less free to run.
But they don't interact well with valgrind. So to also cross check with valgrind we have two different targets.

.devcontainer/Dockerfile
u413-284-si: I meant three RUN cmds. This way we would still execute the other RUN cmds, if one of them fails or am I wrong in this assumption? I think it is in our interest to have the container continue the build process as we could still work with it if not all tools are set up. Many of them are optional like the coverage tool or we can still add them afterwards.

u413-284-si: But closing the conversation as the requested changes are introduced.

u413-284-si (approved):

QCHR1581 (approved):

PR #50, opened 2024-08-30
49 ubsan reference binding to null pointer of type sockaddr

gwolf-011235:
SocketPolicy is a wrapper around socket related functions like getaddrinfo. The wrapper functions had references passed to struct addrinfo. 

In tests MockSocketPolicy is used. In this tests dummy struct addrinfo are created. These structs also have a pointer to a struct sockaddr, which was init to NULL (since not used anyway). But this lead to passing a reference to NULL which was caught by UBSan as undefine behavior.

Changes:
- To avoid generating reference to NULL the functions now take a pointer and have an assert(ptr != NULL)
- The testcases create dummy addrinfo with a dummy sockaddr added, so no NULL pointers are sent


u413-284-si (approved):
Looks good! Nice job on finding all instances where the pointer syntax needed to be adapted.

QCHR1581 (approved):
Good approach to use pointers instead of references to avert undefined behaviour.

PR #51, opened 2024-08-30
45 remove definition of private copy constructors assignment operators

### LINKED ISSUE #45: Remove definition of private copy constructors & assignment operators
To make a class not copyable in C++98 we can _declare_ the copy constructor and copy assignment operator as private in the header file.
If there isn't even a _definition_ in the source file they can't even be invoked in private context. 

There are some definitions (mainly from me) of private copy constructors & assignment operators. If the class should never be copyable these should be removed.

gwolf-011235:
If there isn't even a definition in the source file, private copy constructor and copy assignment operator can't even be invoked in private context.

Fix: removed Copy ctor and assign from
- ALogOutputter.cpp
- Directory.cpp
- Logger.cpp

Still compiles :D

I didn't find any impementations of copy ctor or assign anywhere else, therefore 
Closes #45


inc/Logger.hpp
```diff
@@ -33,11 +33,12 @@ class Logger {
 private:
 	// private constructor to prevent direct object creation
 	Logger();
+	// private destructor to prevent direct object deletion
+	~Logger();
```
u413-284-si, line 37: As you're not performing any manual tasks in the destructor, you can remove this declaration and the definition here and just use the implicit default destructor.

inc/Logger.hpp
```diff
@@ -33,11 +33,12 @@ class Logger {
...
 
 	// disallow copying the instance
 	Logger(const Logger& ref);
```
u413-284-si, line 40: The declarations of copy ctor and assign operator can also be removed. Still compiles.

src/ALogOutputter.cpp
```diff
@@ -18,30 +18,6 @@ ALogOutputter::ALogOutputter() { }
  */
 ALogOutputter::~ALogOutputter() { }
```
u413-284-si, line 19: I think this definition and its declaration in the header can also be removed.

src/Logger.cpp
```diff
@@ -14,32 +14,11 @@ Logger::Logger()
...
- */
-Logger& Logger::operator=(const Logger& ref)
-{
-	if (this == &ref) {
-		return *this;
-	}
-	m_logLevel = ref.getLevel();
-	return *this;
-}
+Logger::~Logger() {}
```
u413-284-si, line 21: see above

u413-284-si (changes requested):
The declarations of the copy ctors and assignment operators can be removed in all classes.

The declaration and definitions of destructors can be removed, if the destructors does not perform any extra steps other than destroying its object.

inc/Logger.hpp
gwolf-011235: Logger is a singleton. Nobody should be able to call the destructor of Logger (// private destructor to prevent direct object deletion).

If it was public, you could call the destructor anywhere (as it's just another function) leading to crash.

inc/Logger.hpp
gwolf-011235: Same thinking as with destructor: Logger is a singleton, nobody should be able to make a copy of it.

src/ALogOutputter.cpp
gwolf-011235: ALogOutPutter is an abstract base class. The destructor is virtual, so that inheriting classes can call their own destructor. Since I declare the destructor I also need to implement it.

So this one is needed

src/Logger.cpp
gwolf-011235: If I remove I get `undefined reference to `weblog::Logger::~Logger()'`

Since I need to call the destructor eventually I also need to implement it. 
The destructor is private for a reason (see above) so I can't use the (public) default version

inc/Logger.hpp
u413-284-si: You mean if the declaration as private is removed and the default destructor is used, this destructor would be public and accessible from anyone?

QCHR1581 (approved):
Nothing additional to note. 

Comments regarding why the destructor of the class Logger needs to be private, the copy ctor and assignment operator need to be declared etc. made it clear what the idea behind those approaches were. 

Good job!

u413-284-si (approved):

PR #61, opened 2024-09-04
58 handle special cases concerning curly brackets

### LINKED ISSUE #58: Handle special cases concerning curly brackets
Following cases regarding curly brackets should be handled (NINX accepts them)

# 1. Curly bracket on the next line

```nginx
server 
{
}
```
Here, the server is not recognized because a bracket is expected to come after the server directive. Therefore, it is not added. No error.

# 2. Whitespaces between server and bracket

```nginx
server         {
}
```
Here, the server is also not recognized because a space is expected to come after server, followed by the bracket. Therefore, it is not added. No error.

# 3. Directive on the same line as the bracket

```nginx
server         {
root /workspaces/webserv/; }
```
Here, an error occurs: "Semicolon missing". The function ``processRemainingLine(line);`` processes a line further if something is found after the semicolon. After only the bracket is filtered out and the loop continues, it checks again if there is a semicolon, which causes the error.

QCHR1581:
# Implemented Fixes

Created implementation for handling following cases concerning curly brackets:

- Curly bracket on the next line
- Whitespaces between server and bracket
- Directive on the same line as the bracket

For details see #58.

---

# Approach for curly brackets on the next line & whitespaces between server and bracket

Both cases are now getting handled by extracting the directive of the current line (in this case ``server``)
and checking if the the directive equals ``server``

```diff filename="ConfigFileParser.cpp"
for (readAndTrimLine(); m_currentLine != "}"; readAndTrimLine()) {
+                 if (getDirective(m_currentLine) == "server") {
-                 if (m_currentLine == "server {") {
			ConfigServer server;
			m_configFile.servers.push_back(server);
			while (m_currentLine != "}") {
				readAndTrimLine();
				readServerConfigLine();
		}
		m_serverIndex++;
	}
}
```

# Approach for a directive on the same line as a bracket

To handle this case, a rewrite of the read server config line loop and the read location config line loop was necessary.
Furthermore the way how to process a remaining line also had to be changed.

## Read server config line loop
```cpp
while (m_currentLine != "}") {
	readAndTrimLine();
	readServerConfigLine();
	}
```

## Read location config line loop
```cpp
while (m_currentLine != "}") {
	readAndTrimLine();
	readLocationConfigLine();
	}
	m_locationIndex++;
	processRemainingLine(line);
	continue;
```
## Process remaining line
```cpp
void ConfigFileParser::processRemainingLine(std::string& line)
{
	size_t semicolonIndex = line.find(';');
	if (semicolonIndex == std::string::npos) {
		line = "";
		m_currentLine = line;
		return;
	}

	line.erase(0, semicolonIndex + 1);
	line = webutils::trimLeadingWhitespaces(line);
	webutils::trimTrailingWhiteSpaces(line);
	m_currentLine = line;
}
```
---

# Reason to change the read loops

The reason to change the approach in this way is to stop the read loops if the current line equals ``}``, even though a directive is on the same line as the ``}``.

When a line like this appears:

```nginx
root /var/www/hmtl; }
```

The directive is read, and the remaining line is processed. This results in the line being reduced to only a ``}`` character. 
Then the m_currentLine is updated, causing the loop to terminate.

Closes #58 

--- 

# Tests

There are tests for all of the mentioned cases to make sure that each of them is handled correctly


gwolf-011235:
Did some tests:
- listen directive behaves strange when on same line:
`server {listen 127.0.0.1:8080;` gives 127.0.0.1:80 (standard values)
Other directives (tested with server_name) seem to work?

- if one server block is added, any additional wrong blocks are ignored
```
server {listen 0.0.0.0:8080;} //<-- closed early
        server_name scifi.com; // from here on everything is ignored (should error)
        root /workspaces/webserv/scifi;
        location / {
            autoindex on;
        }
server {listen 127.0.0.1:8080;} // this new server block is also ignored
```

- if brace is on next line, and directive on same line (combination of the two fixes) it does not like:
```
server //<-- brace on next line 
{listen 0.0.0.0:8080;} // error: Invalid server directive
```

u413-284-si (approved):
Seems fine to me, no additional comments. Nicely done.

gwolf-011235 (approved):
Seems good!
Found test case 3. will be handled in separate branch

PR #63, opened 2024-09-09
Fix makefile asan compilation

gwolf-011235:
### Summary
Small fix for Makefile: better conditional compilation

### Detail
Right now Makefile generates different obj directory depending on target (conditional compilation). This worked on another project, where depending on bonus target different source files where used.
I think it may not be the right fit here: e.g. if we compile name and then test, we need to compile all program objects two times. Also make comp (for creating compile_commands)  is broken, since it generates compile_commands with the wrong flags for the program files.

Changes
- object are now put in subdir default (aka obj/default)
- test has its own CXXFLAGS variable. It makes no difference if test or name is called first, the compile flags are always the same for PROG_OBJS and TEST_OBJS
- new variable SANI controls if sanitizers are enabled: objects are put subdir obj/sani; name has suffix _sani if enabled
- dependencies are always put in BASE_OBJ_DIR (obj) as they don't change if sanitizers are enabled or not
- comp just executes the test target > as flags don't change the correct flags are stored in compile_commands
- added Wdocumentation to the normal flags: it doesn't scream if no Doxygen is there, only if wrong Doxygen

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -31,6 +31,7 @@
 *.out
 *.app
 webserv
+webserv_sani
 unittest
 unittest_sani
 
--- a/Makefile
+++ b/Makefile
@@ -11,7 +11,19 @@
 # **************************************************************************** #
 
 # ******************************
-# *     Verbosity              *
+# *     Targets                *
+# ******************************
+
+NAME := webserv
+
+TEST := unittest
+
+NAME_SANI := $(NAME)_sani
+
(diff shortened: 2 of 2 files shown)


QCHR1581 (approved):
Useful little improvements concerning compilation. 
Looks good to me.

u413-284-si (approved):
Well perceived optimisation opportunity. Good job!

Issue #64, opened 2024-09-15
ResponseStream of ResponseBuilder not cleared of previous response


u413-284-si:
When sending multiple requests to the same server it is witnessed that the received response contains part of the previous response.

This happens when the previous response is longer than the current one. Reason behind this behaviour is that resetStream() in the ResponseBuilder only sets the stream to the beginning. When the new content is shorter than what is already in the stream, only the length of the new content is overwritten and a remainder of the old content stays. This is sent together with new response.

Solution: 
ResponseStream needs to be cleared before a new response is generated.

gwolf-011235:
Are we sure the problem is in ResponseBuilder? 
As explained in #20 the stream is intentionally not cleared. Since we return .c_str() it should only return upto the first zero terminator. 

Did you observe it in connection with CGI? How are you adding the CGI result to the response? Could it be that it doesn't add a zero terminator?

Another place could be in connectionBuildResponse() where the connection buffer gets assigned to the response. The assignment operator should replace the contents of the string with the other string ([operator=](https://en.cppreference.com/w/cpp/string/basic_string/operator%3D)). Maybe its not actually _replacing_ but copying into? 
In #59 the connection buffer gets .clear() always, which crosses out this possibility.

Also: I can't reproduce. I tried with 
- GET request
- a small index.html and 
- a big about.html. 

Switching back and forth always resulted in the correct response being sent. So maybe CGI stuff?

gwolf-011235:
Stringstream seems to cause problems when appending the response body, maybe because of the insertion operator << and how it interacts with appending a std::string.

When the body is a png file, it doesn't work. However if I just switch the stringstream with a std::string, then it works.

I could look into this, but I think it's easier to just use a std::string and do the conversions (mainly in the response headers) with webutils::toString() 

webutils::toString() uses stringstream for conversion. Could be improved by using std::sprintf, if we have performance issues: see https://en.cppreference.com/w/cpp/string/basic_string/to_string for c++11 std::to_string() which works "as if using sprintf". However CPP Core Guidelines really don't like sprintf as it is a va_args function

PR #65, opened 2024-09-17
Feat static website

gwolf-011235:
### Summary

Adds a static website to test functionality. Nothing pretty but functional. Can be extended as needed.

### Detail 

- All files are located in html subfolder
- Static website consists of
  - index.html: has links to some subsites
  - content.html: just some basic html content with formatting
  - styled.html: random content, but this also has a stylesheet embedded which needs to be fetched
  - images.html: has images in different formats (jpg, png, webp) and one external
  - run_cgi.html: can be used to test cgi scripts, right nowhas just dummy forms
- there is also a directory subfolder to test file index generation
- also has a [favicon](https://en.wikipedia.org/wiki/Favicon)

### Misc

- createDummyConfig() adapted to support this static website. When parser is ready we should create a configfile which mimics dummyConfig
- adapted one test to clear all location > prev just popped one. Since dummyConfig() generates server with two locations this lead to a "fail" of the test. Now fixed.


u413-284-si (commented):
Webpage works in general as declared except for image display, but this issue is already being addressed in #64 .

Favicon not working due to same reason as the images I suppose.

Regarding the CGI page, to make the links work the according locations need to be set in the ConfigFile. This can be done when both this PR and the CGI PR are merged.

u413-284-si (approved):

QCHR1581 (approved):
Great to see how basic pages can be displayed by using the webserver!

Displaying images and running cgi scrips is in work as Sirui already mentioned. 

Otherwise had troubles to connect to the server with a browser outside of the devcontainer, but this was my own issue :)

PR #67, opened 2024-09-18
64 responsestream of responsebuilder not cleared of previous response

gwolf-011235:
### Summary

1. Fixes issue with binary data (e.g. image.png) being corrupted when sent in response
2. Removes unnecessary optimization
3. Prepares for variable sized clientHeader / Body buffer
4. Removes m_receivedBytes from connection

### 1. Binary data

- ResponseBuilder has a static stringstream which contains response header and body
- Binary data was corrupted probably because body was added via << operator to the stream
- solution: only use stringstream for headers, as only for them conversions are needed (e.g. size for content-size)
- image.png can now be sent.
- cgi should not need zero terminator anymore

### 2 Unnecessary optimization

- Stringstream is reused, so that it's only constructed once
- It was also used in a unnecessary efficient way by using a c_str() "hack" (see #20 for more details)
- now just clear the stream like normal people to avoid additional problems
- if really needed optimization can be put back in    

### 3. Prepare for variable sized buffers

- we talked about using std::string as buffer may not be optimal as the container is made for c strings (with zero terminator as delimiter). For example `std::string += char[]` invokes a strlen() for the char array which is problematic for binary data
- more "appropriate" would be std::vector< char >; looked at it, would prob require a bigger rewrite
- left connection buffer as std::string 
- changed local buffer to use a std::vector< char >, so it can be resized (variable array is an C99 extension). At first wanted to also use std::string. But .clear() is not guarenteed to not reallocate
- use explicit .append(start, size) method to append to connection buffer

### 4. Remove receivedBytes

- class Connection had variable receivedBytes
- Wasn't really used, as receivedBytes is the same as buffer.size(). Therefore removed.

### DIFF
--- a/inc/ResponseBuilder.hpp
+++ b/inc/ResponseBuilder.hpp
@@ -11,9 +11,9 @@
  * @brief Class to build a HTTP response.
  *
  * This class is responsible to build a HTTP response based on the request received.
- * The ConfigFile is used to get the configuration of the server.
- * It uses the classes TargetResourceHandler to handle the target resource and
- * ResponseBodyHandler for the body of the response.
+ * It uses a stringstream m_responseHeader to convert variables to string. The stringstream is reused.
+ * It uses the class ResponseBodyHandler to construct the body of the response.
+ * Each time a new response is built the old one is overwritten.
  * The FileSystemPolicy passed in the constructor needs to outlive this class. It is passed to subclasses.
  * A mock of FileSystemPolicy can be used for testing.
  */
@@ -29,11 +29,10 @@ private:
 	void appendHeaders(const HTTPRequest& request);
 	std::string getMIMEType(const std::string& extension);
 	void initMIMETypes();
-	void resetStream();
+	void resetBuilder();
 
 	std::map<std::string, std::string> m_mimeTypes;
 	const FileSystemPolicy& m_fileSystemPolicy;
-	std::stringstream m_responseStream;
+	std::stringstream m_responseHeader;
 	std::string m_responseBody;
-	bool m_isFirstTime;
 };
--- a/inc/Server.hpp
+++ b/inc/Server.hpp
@@ -63,6 +63,8 @@ public:
 	const std::map<int, Connection>& getConnections() const;
 	const std::vector<ConfigServer>& getServerConfigs() const;
 	time_t getClientTimeout() const;
+	std::vector<char>& getClientHeaderBuffer();
+	std::vector<char>& getClientBodyBuffer();
 
 	// Setters
 	bool registerVirtualServer(int serverFd, const Socket& serverSock);
@@ -105,6 +107,8 @@ private:
 	time_t m_clientTimeout; /**< Timeout for a Connection in seconds */
 	std::map<int, Socket> m_virtualServers; /**< Listening sockets of virtual servers */
 	std::map<int, Connection> m_connections; /**< Current active Connections */
+	std::vector<char> m_clientHeaderBuffer; /**< Buffer for reading request header */
+	std::vector<char> m_clientBodyBuffer; /**< Buffer for reading request body */
 	RequestParser m_requestParser; /**< Handles parsing of request */
 	FileSystemPolicy m_fileSystemPolicy; /**< Handles functions for file system manipulation */
 	ResponseBuilder m_responseBuilder; /**< Handles building of response */
--- a/src/ResponseBuilder.cpp
+++ b/src/ResponseBuilder.cpp
@@ -16,7 +16,7 @@ ResponseBuilder::ResponseBuilder(const FileSystemPolicy& fileSystemPolicy)
  *
  * If the response body is empty, only the response header is returned.
  * Otherwise, the response header and body are returned.
- * The response doesn't chnage until the next buildResponse() call.
+ * The response doesn't change until the next buildResponse() call.
  * @return std::string Response.
  */
 std::string ResponseBuilder::getResponse() const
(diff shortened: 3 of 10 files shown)


src/ResponseBuilder.cpp
```diff
@@ -7,102 +7,107 @@
...
 /**
  * @brief Get the built response.
  *
- * The stream is returned as c_str() since it is reused.
- * The buffer is never cleared all the way, only reset to the beginning.
- * If str() would be returned, old parts of the stream would be included.
- * This works since every thing inserted into the string sets a new zero terminator.
+ * If the response body is empty, only the response header is returned.
+ * Otherwise, the response header and body are returned.
+ * The response doesn't chnage until the next buildResponse() call.
```
u413-284-si, line 19: typo: change

inc/ResponseBuilder.hpp
```diff
@@ -29,11 +29,10 @@ class ResponseBuilder {
...
 	std::string getMIMEType(const std::string& extension);
 	void initMIMETypes();
-	void resetStream();
+	void resetBuilder();
 
 	std::map<std::string, std::string> m_mimeTypes;
 	const FileSystemPolicy& m_fileSystemPolicy;
-	std::stringstream m_responseStream;
+	std::stringstream m_responseHeader;
 	std::string m_responseBody;
```
u413-284-si, line 37: So the std::string works for binary data if the << and += operators are avoided?

inc/Server.hpp
```diff
@@ -105,6 +107,8 @@ class Server {
 	time_t m_clientTimeout; /**< Timeout for a Connection in seconds */
 	std::map<int, Socket> m_virtualServers; /**< Listening sockets of virtual servers */
 	std::map<int, Connection> m_connections; /**< Current active Connections */
+	std::vector<char> m_clientHeaderBuffer; /**< Buffer for reading request header */
```
u413-284-si, line 110: Why have it as a member instead of just a local variable as before the char[]?

src/Server.cpp
```diff
@@ -893,6 +918,12 @@ void connectionReceiveBody(Server& server, int clientFd, Connection& connection)
 				connection.m_status = Connection::BuildResponse;
 				server.modifyEvent(clientFd, EPOLLOUT);
 			}
+			else if (connection.m_buffer.size() == Server::s_clientMaxBodySize) {
```
u413-284-si, line 921: Why moved down here? The check should be performed directly after the connection.m_buffer has been updated with the new read data.

u413-284-si (changes requested):
So do we still switch from std::string to std::vector<char> (or std::vector<unsigned char> as proposed [here](https://stackoverflow.com/questions/7064822/how-to-push-binary-data-into-stdstring))? Question is, is that switch necessary to avoid errors in transport/conversion of data or are we good with the status quo. I can't tell yet. Would need to be tested/verified after the current branches are merged.

Other than that some design questions raised by changes introduced in this branch.

src/ResponseBuilder.cpp
gwolf-011235: Fixed

inc/ResponseBuilder.hpp
gwolf-011235: The root problem is that binary data doesn't work like c-strings: they can contain "zero terminators" aka a zero byte in the middle of the data.

std::string acts like a std::vector for char (bytes) but for compatibility for "old" c-strings also has various functions which work on a const char * (which is expected to be zero terminated). This can lead to bugs if invoked on data which holds non-string data (binary data). One has to be careful not to invoke the wrong function. 
So std::string works but actually "isn't the right tool".

Std::stringstream has the same "problem": it's a stream class to operate on strings. The << operator works on arithmetic types (const char *) and not on std::string. Which leads to the same problem with binary data. We could work around this by constructing a stream out of the std::string and insert the stream buffer (which should insert it "as is").

So both containers work with binary data, but you have to be careful to not invoke functions which work on const char *.

I chose to use stringstream for response header as it is text = string.
I kept std::string to keep consistent with other parts of our code.
I think std::vector<unsigned char> would be the "correct" container to choose when handling binary data.

inc/Server.hpp
gwolf-011235: The char[] can be initiated on the stack. 

In C++98 you can't initiate a std::vector on the stack, it always allocates on the heap. To avoid allocations everytime we enter this function I created member variable

src/Server.cpp
gwolf-011235: More closely mimicked connectionReceiveHeader: we track with bytesToRead how many bytes are left to read. 

Before it would always read the full amount of bytes irregardless of how many have already been read. If 1mb is allowed and we already received 900k, we would read another 1mb.
With bytesToRead we check beforehand and only read whats remaining. 

Then we check if we have a complete body. If we don't have complete body, and already read maximum bytes we know that we can never receive the body - all allowed bytes are used. Therefore we can send error

inc/ResponseBuilder.hpp
u413-284-si: Thanks for the clarification. Do I understand you correctly that ftm we stay with the std::string and try to avoid functions that utilise a const char* and have no intention of switching to std::vector yet (except for it becomes really unavoidable)?

src/Server.cpp
u413-284-si: Ah yes, I see. Because you added that limit on how many bytes are read, it could be in certain cases that the max. body size has been reached and at the same time the body is complete. If we were to check for the max size before, you would discard a valid request.

inc/ResponseBuilder.hpp
gwolf-011235: Yes, I would leave std::string where it's already in codebase and works. But we should keep an eye out for unintentional usages for const char *

If we add new buffers, I propose we prefer std::vector - though there shouldn't be much "opportunities" left.

u413-284-si (approved):

QCHR1581 (approved):
No questions or additional comments from my side. 

Tested an image (of course a picture of a cat) and it worked out well!

PR #73, opened 2024-09-20
70 require host header in each request

### LINKED ISSUE #70: Require Host header in each request
Add check after header parsing

u413-284-si:
## Overview
1. Check for missing host header
2. Check for empty host header value
3. Check for multiple host headers in request
4. Determine if host value is ip (w or w/o port) or a hostname
5. Check for valid ip (w or w/o port) 
6. Check for valid hostname
7. Code restructure
8. Check for not allowed method
9. Tests

### re 1.
- simple check if Host header is present in std::map<std::string, std::string> headers, if not throw

### re 2.
- simple check if the value of the Host header is empty, if yes throw

### re 3.
- during parsing of the respective headers, check if there is more than one host header in the request, if yes throw

### re 4.
- Host value can be either a hostname or an ip address (with or without a port)
- implemented check to differentiate

### re 5.
- use Chris' methods to check for valid ip address and/or port
- therefore moved isIpAddressValid() and isPortValid() to utilities as it is shared

### re 6.
- if is not an ip address, then see if it is a valid hostname
- criteria follow RFC 1035, 1123 and 3986
- detailed description in code documentation

### re 7.
- as mentioned moved Chris' methods
- restructured RequestParser to align with header file and its structure

### re 8.
- create new method to check for allowed methods if there is a body
- set httpStatus to StatusMethodNotAllowed instead of general StatusBadRequest -> general check should be performed to specify status code

### re 9.
- add new unittests to test various valid and invalid host headers

closes #70


inc/RequestParser.hpp
```diff
@@ -20,12 +20,15 @@
...
 
-const int decimalBase = 10;
+const int g_decimalBase = 10;
 
 /* ====== CLASS DECLARATION ====== */
 
 class RequestParser {
 public:
+	static const int s_maxLabelLength = 63; /**< Maximum length for labels (the parts between dots in a domain name)  */
+	static const int s_maxHostNameLength = 255; /**< Maximum length for hostname */
```
gwolf-011235, line 30: I think a webserver would mostly deal with hostnames used in DNS. Therefore I would limit to 253

From [Stackoverflow](https://stackoverflow.com/questions/8724954/what-is-the-maximum-number-of-characters-for-a-host-name-in-unix/28918017#28918017): "Hostnames used in DNS can be as long as 253 bytes as a fully qualified domain name" 

Or the linked [MSDN blog article ](https://devblogs.microsoft.com/oldnewthing/20120412-00/?p=7873) with useful illustration 
(in here it's also mentioned that with utf-8 we would need to calculate differently - let's just ignore that :D )

inc/error.hpp
```diff
@@ -19,6 +19,12 @@
 // HTTP REQUEST HEADER ERRORS
 #define ERR_HEADER_COLON_WHITESPACE "Invalid HTTP request: Whitespace between header field-name and colon detected"
 #define ERR_HEADER_NAME_INVALID_CHAR "Invalid HTTP request: Invalid char in header field name"
+#define ERR_MISSING_HOST_HEADER "Invalid HTTP request: Missing host header"
+#define ERR_EMPTY_HOST_VALUE "Invalid HTTP request: Missing hostname"
+#define ERR_MULTIPLE_HOST_HEADERS "Invalid HTTP request: Multiple host headers"
+#define ERR_INVALID_HOSTNAME "Invalid HTTP request: Invalid hostname"
+#define ERR_INVALID_IP "Invalid HTTP request: Invalid IP"
+#define ERR_INVALID_IP_WITH_PORT "Invalid HTTP request: Invalid IP with port"
```
gwolf-011235, line 27: I would only use the first error message as it bring the point accross.

If you want to keep all three I would at least specify that the invalid IP (with port) was determined in context of hostname

src/RequestParser.cpp
```diff
@@ -817,3 +925,125 @@ bool RequestParser::isMethodAllowedToHaveBody(HTTPRequest& request)
...
+ * @brief Checks if the given host value is a valid IP address.
+ *
+ * This function parses the provided host value to determine if it is a valid
+ * IPv4 address. It also checks if the IP address includes a port number.
+ *
+ * @param hostvalue The host value to be checked.
+ * @param hasPort A reference to a boolean that will be set to true if the IP address includes a port number, false otherwise.
+ * @return true if the host value is a valid IP address, false otherwise.
+ */
+bool RequestParser::isIPAddress(const std::string& hostvalue, bool& hasPort)
```
gwolf-011235, line 1017: This function does a whole lot, which is also covered by isIpAddressValid() and isPortValid() 
(it even says so in doxygen: "to determine if it is a valid IPv4 address").

I wouldn't even differentiate between IP and hostname in error message.

If you want to differentiate, I don't think you need this function. In validateHostHeader()
- check if you find a colon ":" --> if yes it has to be IP:port
- check if (ipAddress.find_first_not_of("0123456789.") != std::string::npos) --> has to be IP
- else has to be hostname

src/RequestParser.cpp
```diff
@@ -788,6 +608,294 @@ void RequestParser::validateTransferEncoding(HTTPRequest& request)
...
+ *
+ * @param headerName The name of the header to validate.
+ * @param request The HTTP request object containing the headers.
+ *
+ * @throws std::runtime_error if multiple "Host" headers are found.
+ */
+void RequestParser::validateNoMultipleHostHeaders(const std::string& headerName, HTTPRequest& request)
+{
+	if (headerName == "Host") {
+		if (request.headers.find("Host") != request.headers.end()) {
```
gwolf-011235, line 699: Field names are case-insensitive per RFC 9110 - 5.1 [Field Names](https://www.rfc-editor.org/rfc/rfc9110.html#section-5.1-3)

This means if a client sends two headers with "Host = [...]" and "HOST = [...]" it wouldn't be caught.

General: I think we should sanitize the field names, e.g. all lowercase

gwolf-011235 (changes requested):
Looks good. I like the usage of existing functions. Also nice amount of tests!

I think we have to think about how we store headers, as they are case insensitive > maybe separate issue?

inc/RequestParser.hpp
```diff
@@ -20,12 +20,15 @@
 
 /* ====== DEFINITIONS ====== */
 
-const int decimalBase = 10;
+const int g_decimalBase = 10;
```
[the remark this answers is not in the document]
u413-284-si: As it is used in several files, I would argue that it makes sense to have it global and that it would be beneficial to declare it in one central .cpp file as external const and then use a header which can be included wherever needed following the advice [here](https://www.learncpp.com/cpp-tutorial/sharing-global-constants-across-multiple-files-using-inline-variables/). This way it could theoretically be changed in one place and take effect in all of the code instead of doing it manually everywhere, even when in this case it is unlikely to be changed.

src/RequestParser.cpp
```diff
@@ -1,196 +1,67 @@
 #include "RequestParser.hpp"
+#include "StatusCode.hpp"
+#include "error.hpp"
+#include "utilities.hpp"
+#include <cstddef>
+#include <iostream>
+#include <map>
+#include <vector>
```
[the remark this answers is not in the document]
u413-284-si: Removed.

src/RequestParser.cpp
```diff
@@ -788,6 +608,294 @@ void RequestParser::validateTransferEncoding(HTTPRequest& request)
 	}
 }
 
+/**
+ * @brief Validates if the HTTP request method is allowed to have a body.
+ *
+ * This function checks if the HTTP request contains a body and whether the
+ * method used in the request is allowed to have a body. If the request has a
+ * body but the method is not allowed to have one, it sets the HTTP status to
+ * Bad Request and throws a runtime error.
```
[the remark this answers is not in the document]
u413-284-si: Updated.

src/RequestParser.cpp
```diff
@@ -788,6 +608,294 @@ void RequestParser::validateTransferEncoding(HTTPRequest& request)
...
+	std::map<std::string, std::string>::const_iterator iter = request.headers.find("Host");
+	if (iter == request.headers.end()) {
+		request.httpStatus = StatusBadRequest;
+		throw std::runtime_error(ERR_MISSING_HOST_HEADER);
+	}
+
+	if (iter->second.empty()) {
+		request.httpStatus = StatusBadRequest;
+		throw std::runtime_error(ERR_EMPTY_HOST_VALUE);
+	}
```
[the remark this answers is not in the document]
u413-284-si: That is a good question. 

I think the sentence in the RFC refers to CONNECT requests which we don't have to handle. These kind of requests are the only once that use the [authority-form](https://www.rfc-editor.org/rfc/rfc9112.html#section-3.2.3). Therefore we would not encounter requests encompassing an authority component. Any rules that pertain to that should then not apply to us (?)

As the server_name directive is optional, we would then default to using the machine's hostname or IP address for that server block. But I think that still every request should have a host field value, and if no server block matches this value, then the default server is used.

inc/RequestParser.hpp
u413-284-si: Really nice article, very neatly explained. Changed to 253!

inc/error.hpp
u413-284-si: It mostly helped me with debugging. Specified to hostname context.

src/RequestParser.cpp
u413-284-si: Very good suggestion, as such I like it. But how do you handle these special cases:

InvalidHostnameLikeIP -> Host: 377.3.1.9999
InvalidHostnameLikeIP2 -> Host: 127.3.43.1.1

These look like ip addresses, but are not and should be handled by the isValidHostname() function. The isIpAddressValid() function does not recognize the error as it only checks of the first three segments. 

Idea: let the isIpAddressValid() handle these cases and update it with the code of the isIPAddress() function:

```
std::vector<std::string> segments;
	size_t segmentStart = 0;
	size_t segmentEnd = 0;

	while (segmentEnd != std::string::npos) {
		segmentEnd = hostvalue.find('.', segmentStart);
		std::string segment = (segmentEnd == std::string::npos)
			? hostvalue.substr(segmentStart)
			: hostvalue.substr(segmentStart, segmentEnd - segmentStart);
		segments.push_back(segment);
		segmentStart = segmentEnd + 1;
	}

if (segments.size() != 4)
		return false;
for (std::vector<std::string>::const_iterator citer = segments.begin(); citer != segments.end(); ++citer) {
		if (citer->empty() || citer->size() > 3)
			return false;
}
... then the existing code to check if is larger than 255 or lower than 0.
```

src/RequestParser.cpp
u413-284-si: Good finding. Then I suggest while parsing the headers we send all field names through tolower() - or more precisely all of the letters - and then do the other checks.

src/RequestParser.cpp
u413-284-si: I rewrote the isIPAddressValid() to incorporate the check for amount of segments and segment size:
```
bool isIpAddressValid(const std::string& ipAddress)
{
	if (ipAddress.find_first_not_of("0123456789.") != std::string::npos)
		return false;

	std::vector<std::string> segments;
	size_t segmentStart = 0;
	size_t segmentEnd = 0;

	while (segmentEnd != std::string::npos) {
		segmentEnd = ipAddress.find('.', segmentStart);
		std::string segment = (segmentEnd == std::string::npos)
			? ipAddress.substr(segmentStart)
			: ipAddress.substr(segmentStart, segmentEnd - segmentStart);
		segments.push_back(segment);
		segmentStart = segmentEnd + 1;
	}
	if (segments.size() != 4)
		return false;

	const short maxIpValue = 255;
	const short minIpValue = 0;
	const int base = 10;
	for (std::vector<std::string>::const_iterator citer = segments.begin(); citer != segments.end(); ++citer) {
		if (citer->empty() || citer->size() > 3)
			return false;
		const long segmentValue = std::strtol(citer->c_str(), NULL, base);
		if (segmentValue > maxIpValue || segmentValue < minIpValue)
			return false;
	}
	return true;
}
```

src/RequestParser.cpp
u413-284-si: Added function in utilities to lowercase the parsed field name.

src/RequestParser.cpp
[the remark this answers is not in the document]
gwolf-011235: Ok makes sense

inc/RequestParser.hpp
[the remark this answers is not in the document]
gwolf-011235: Honestly: I don't like it. For me the raison d'etre of this variable (as global or local) is to avoid a magic number. Having a local const int fullfils this purpose. 

Using a global seems not optimal to me for the following reasons:
1. introduces indirection: to fully understand the code you have to refer to another file 
2. increases compile/link time: we add two files
3. adds dependency: when the files change, all dependent files need to be recompiled
4. hard enforceability: if this global is not known to a contributor, it won't be used.

OTOH all of these points are mostly negligible. I'm fine with it!

If you want to add the two files, then I would also change the existing const local ints to use the global. I only found two in 
- isIpAddressValid()
- isPortValid()

src/RequestParser.cpp
gwolf-011235: I would use your split() here:

`const std::vector<std::string> segments = webutils::split(hostvalue, ".");`

Maybe we should also add some tests? If I understood correctly an IP in form 127.3.43.1.1 would have been accepted as correct?

src/RequestParser.cpp
gwolf-011235: Maybe we should also track this as issue for existing stuff?

I found in handleCompleteRequestHeader() a search for "Host", maybe there are more

inc/RequestParser.hpp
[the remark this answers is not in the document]
gwolf-011235: Yeah, but like you said the constant won't really change (else wouldn't be much of a constant).

A more realistic change would be the addition of a new constant, which would change included header. This would need a recompilation of all dependent .cpp (at least with make as build system), even if the new constant is not used.

Yeah go for it!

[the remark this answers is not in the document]
u413-284-si: Implemented and works!

src/RequestParser.cpp
u413-284-si: Ah yes, very good hint! Changed. There are two tests in test_parseHeader_Headers.cpp, namely the aforementioned:

InvalidHostnameLikeIP
InvalidHostnameLikeIP2

src/RequestParser.cpp
u413-284-si: Changed to "host" in handleCompleteRequestHeader()

gwolf-011235 (approved):
Looks good to me! 

Little things could be done before merge, but not that important:
- utilities.cpp has automatic header includes
- constant{.cpp, .hpp} don't seem to be correctly formatted > both have spaces after namespace.

u413-284-si:
> Looks good to me!
> 
> Little things could be done before merge, but not that important:
> 
>     * utilities.cpp has automatic header includes
> 
>     * constant{.cpp, .hpp} don't seem to be correctly formatted > both have spaces after namespace.

Thanks, removed the headers. Don't understand the formatting point though, what space? The formatting seems to me equal to the webutils one.

inc/constants.hpp
```diff
@@ -0,0 +1,10 @@
+#pragma once
+
+/* ====== LIBRARIES ====== */
+
+/* ====== DECLARATIONS ====== */
+
+namespace constants {
+    extern const int g_decimalBase;
```
gwolf-011235, line 8: Here is 4 spaces after namespace opening, indenting the global once

But it shouldn't have an indent: 
in .clang-format the setting is `NamespaceIndentation: Inner`
This causes inner namespaces (a namespace in a namespace) to be indented, see [documentation](https://clang.llvm.org/docs/ClangFormatStyleOptions.html#namespaceindentation)
Since we only have one namespace constants it should be formatted to 
```
namespace constants {
// i like an extra line here
const int g_decimalBase = 10;

} // constants
```
This is also what I get when I autoformat the file (except the extra line after namespace)

See also utilities{.hpp, .cpp} where its formatted the same way

u413-284-si: Oh, inside the namespace in front of its members. Thought you meant after the word "namespace" itself. Changed.

QCHR1581 (changes requested):
Looks good overall. Also nice idea to use the already implemented checks for an ip address (while improving this function) and a port. 

Only got a question regarding the initialization of two member vars.

src/RequestParser.cpp
```diff
@@ -1,196 +1,60 @@
...
+ * @brief Constructs a new RequestParser object.
  *
- * Only checks for the carriage return \r, as the linefeed \\n is discarded
- * by std::getline
- * @param str	Input string to be checked.
- * @param request The HTTP request object to be filled.
- * @throws std::runtime_error If the input string does not contain a single
- *         carriage return, an error is thrown and the error code is set to StatusBadRequest.
+ * This constructor initializes the RequestParser object with default values.
+ * The `m_hasBody` member is set to `false` and the `m_chunked` member is set to `false`.
```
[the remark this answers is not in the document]
u413-284-si: The comment was referring to a previous version of the RequestParser when it still had these two members. They are moved into the HTTPRequest now and are initialised with its ctor. 

The comment has been updated.

QCHR1581 (approved):

Issue #30, opened 2024-07-29
Add default location when creating a server


QCHR1581:
When a server gets created a default location gets added with the path "/". 

If a location is specified in the config file with the path "/" the default location values get changed accordingly to the ones provided .

PR #83, opened 2024-09-27
78 reject not allowed method

gwolf-011235:
- Add simple check if method is allowed 

In `handleCompleteRequestHeader()` after location has been found. Since we have bool array we can just use method enum. If not allowed sets status to `StatusMethodNotAllowed`. Also added test for `handleCompleteRequestHeader()`

This only works if #30 is implemented: a connection should always have an assigned server (default server), it should also have a default location (default location)

- misc: formatting

Saved Server.cpp and it made some changes to formatting -> function one-liners into more-liners


u413-284-si (approved):
Passt

QCHR1581 (approved):
Looks good to me. 

#30 will be implemented to make this properly work.

PR #90, opened 2024-10-31
89 sending multiple requests on same connection not processed correctly

### LINKED ISSUE #77: Lowercase header field names
Header field names are saved in lower case in the std::map<std::string, std::string> headers. Any usage of the map by invoking a field name shall be done in lower case.

### LINKED ISSUE #89: Sending multiple requests on same Connection not processed correctly
First request is processed as expected and delivers the  correct response.

Following requests trigger a bad request response

u413-284-si:
## Overview
After merge with main the request headers were all stored in lower case. This caused errors throughout the code where the headers were searched for with initial upper case letters, eventually resulting in a broken chain of processing multiple requests involving a CGI POST request.

All instances invoking the headers map are performed now in lower case. The observed behaviour has resumed expectations allowing multiple requests be sent and processed correctly.

closes #89 
closes #77

### DIFF
--- a/src/CGIHandler.cpp
+++ b/src/CGIHandler.cpp
@@ -58,10 +58,10 @@ CGIHandler::CGIHandler(Connection& connection, const ProcessOps& processOps)
 
 	/* ========= Set up environment for CGI script ========= */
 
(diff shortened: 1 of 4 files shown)


QCHR1581 (approved):
Simple changes made, nothing to add.

PR #91, opened 2024-11-01
33 always close connection if a bad request 400 is detected

### LINKED ISSUE #33: Always close connection if a BAD REQUEST (400) is detected
extracted from RFC 9112:
"When a server listening only for HTTP request messages, or processing what appears from the
start-line to be an HTTP request message, receives a sequence of octets that does not match the
HTTP-message grammar aside from the robustness exceptions listed above, the server SHOULD
respond with a 400 (Bad Request) response and close the connection."

TO DOs:

- remove boolean in HTTPRequest struct
- implement closing of connection by the server

### LINKED ISSUE #74: Check for specific status code application
Identify if there are certain status codes which are more appropriate than the general 400 BadRequest which is widely applied as of now

u413-284-si:
## Overview

Sets flag _shallCloseConnection_ to true whenever the received HTTP request does not correspond to the message grammar set in RFC 9112 and the server is supposed to respond with a 400 Bad request before closing the connection. Added header _Connection_ in the response: if _shallCloseConnection = true_, then Connection: close is sent, otherwise Connection: keep-alive.

Tests adapted to reflect the boolean being set.

To verify in which cases the connection shall be closed, an examination of the various response codes was also undertaken. No deviation from the current implementation was found.

closes #33 
closes #74

### DIFF
--- a/inc/ResponseBuilder.hpp
+++ b/inc/ResponseBuilder.hpp
@@ -28,7 +28,7 @@ public:
 private:
 	void appendStatusLine(const HTTPRequest& request);
 	void appendHeaders(const HTTPRequest& request);
-	void appendHeadersCGI();
+	void appendHeadersCGI(const HTTPRequest& request);
 	std::string getMIMEType(const std::string& extension);
 	void initMIMETypes();
 	void resetBuilder();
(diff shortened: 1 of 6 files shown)


QCHR1581 (approved):
Changes were correctly done according to RFC 9112. 
Also checked our responses compared to NGINX and they were the same. 

Gj!

PR #102, opened 2024-12-09
97 targetresourcehandler throw if no default location

gwolf-011235:
[Updated]
After discussion throw was deemed unnecessary. 
Changes are now:
- matchLocation() sets to .begin()
- check removed in locateTargetResource()
- replaced assert() with statusCode to if (is not in range) > set to Internal Server Error

[Old]
Beforehand if no location match was found the status code was set to internal server error. This led to a bug when ResponseBodyHandler tried to locate a custom error page, by dereferencing the not set location.

Now TargetResourceHandler throws. This is only catched in main, meaning the server is down.

The case that no location is found **should not happen** since we have a default location of "/" and RequestParser makes sure that the path starts with "/".


u413-284-si (commented):
I am bit torn due to the fact that **IF** it throws, our server would be down:

One the one hand, this should never happen as we always have the default location as activeLocation. On the other hand it states in the subject in a fat red box:

> The important thing is resilience. Your server should never die.

So, I wonder if we should catch the throw before the main,  in handleCompleteRequest() for example, just to make absolutely sure that the server simply can't die because of that. In case we forgot about some scenario where it still throws. I mean it is not too complicated to catch in advance.

gwolf-011235:
I thought so that it would feel iffy.

Since our precondition is that path should always start with / and we have a default location the case were this throws would be 
1. the stars aligned and shifted some bits in the path string
2. we made an error

In both cases we are in an unexpected state. 

A similar case would be the assert in the status code <<operator overload. This would also kill the server in both cases (e.g. we made a parsing error and have an undefined status code floating around)

We could catch. But then we could also just leave it like it is and make extra check in error body (to avoid segfault > just use default errorpage)

QCHR1581:
I kind of understand both sides from you two. 

In my opinion, it's crucial that we prevent our server from shutting down.

We currently catch exceptions before the main entry point for invalid inputs, such as incorrect config files or requests and fails of internal functions. 

This specific case is neither of both. The user is not able to provide a location path, that we can not match with the default location ``/`` and this is also not an internal function. 

So therefore we need another approach for this situation in my opinion. 

Is there another way, to set the status to an internal server error without throwing an exception?

u413-284-si:
Shouldn't we be sending the error page for 404 Not Found in case no location was matched? And in addition to that a check in the handleErrorBody() function to prevent dereferencing a non existing location would be the alternative Chris is asking about. So yes, we can just pursue this approach instead of the throw.

gwolf-011235:
Tl;dr at the end!

> The user is not able to provide a location path, that we can not match with the default location / and this is also not an internal function.

> Shouldn't we be sending the error page for 404 Not Found in case no location was matched?

Maybe I missunderstood, but I think we are not on the same page here.

Location matching happens after the request headers (i.e. request line and headers) are parsed. At this moment we should have the following state:
1. the active server config has at least one location with path "/".
When a new server config is created a location is added with "/" (default location). If the ConfigParser doesn't have an error (e.g. for whatever reason it overwrites the path unintentionally) this default location should exist.
2. this also means that at least one location is at locations.begin()
For this location to not exist, the ConfigFileParser would have to unintentionally delete it 
3. the request.uri.path starts with a "/"
RequestParser makes sure that the uri starts with a "/". If the RequestParser doesn't have an error (e.g. it lets something through which has not "/" at beginning) the path should start with "/".

Therefore Location matching would always match, since "/" matches with "/[...]" guaranteed by (1) and (3). It doesn't match in the exceptional case that our own defined preconditions are not there (which for me is a serious dent in the logic)

> But what happens in case of RequestParser rejects a request because the path doesn't start with "/"?

When a connection is created its server config is set to the first matching server for host:port. Its location variable is init to locations.begin() of the server config. (2) ensure that locations.begin() exists
Since RequestParser says no, it doesn't search for the correct location. It also shouldn't go boom in createErrorPage() because it is at locations.begin(), which should exist. 
However it does search for a custom error_page, if location.begin() has an error_page defined. The search path then is the string of error_page. ConfigFileParser will make sure, that string of error_page directive starts with "/". So we end up in the same situation as above: match "/" with "/[...]" guaranteed by (1) and (3)

> Then why have the throw at all if it shouldn't happen anyway?

I thought about it, it doesn't make much sense. It would only be active in the exceptional case that our logic has a serious dent. For me it was more of a temporary coding guard instead of a finished line. Which didn't even work, because it was not in place when it was needed.

> But there was a segfault!

Yes. But the program was in a state where we already knew that something was missing: default location. A configfile which passes through the ConfigFileParser has default location. Since its creation is in the constructor its also not that easy to create a configfile without default location for testing.

> It still feels wrong

Suggestion: location matching starts at locations.begin() instead of locations.end(). If nothing matches (for whatever reason) we are still at locations.begin() which should exist (2.)
This would also make the check for if location exist in error body not needed.

> Other consideration 

Since we want to keep our server alive as long as possible we may want to rethink the assert() lines above the switch cases. Especially for enum statusCode. Status Code will be parsed for return directive, which means there could be a parsing error (e.g. a value of 5000) which would kill Server as soon as it tries to use it (e.g. for response line).

Tl;dr:
- Remove the throw line (with no replacement)
- Return locations.begin() in matchLocation() if no match found
- Rethink assert() for statusCode switch cases

gwolf-011235:
> Why was the locationMatch set to locations.end() to begin with?

Early in development I thought that nginx would could also not match a location. So I set it to end() and if no found InternalServerError. 
But after research nginx behavior is more like having a "/" as catch all. In the comment I even had "remove if default location is defined", so at that moment I even thought of it as unnecessary. 
I think I just thought if the check hits we should throw. But as written above: it doesn't make much sense to even check.

I updated like suggested

u413-284-si (approved):

QCHR1581 (approved):
Well thought-out solution!

It makes a lot of sense to return ``locations.begin()`` if no match can be found for a location, which also prevents throwing.

PR #103, opened 2024-12-11
96 implement directive alias

gwolf-011235:
### Overview

Implements directive alias - [nginx docs](http://nginx.org/en/docs/http/ngx_http_core_module.html#alias)
1. New function replaceAlias()
2. New variable std::pair<bool, std::string> alias and check for it
3. Added tests and updated test index page with link
4. Not implemented: parsing of alias
 
#### 1. New function replaceAlias()
Searches `input` from the beginning for `alias`. If found replaces `alias` with `replacement`. 
If a location has alias set, the location path gets replaced by the alias. Since a location match only happens when the location path matches the request path from the beginning, we can assume that we find the alias at the beginning. 

#### 2. New variable std::pair<bool, std::string> alias and check for it
Alias variable is init to false, empty string. 
After we match a location we check if alias is set by checking the bool. If it is true we replace alias. If false we append to the root. This means its either root or alias (nginx also accepts only one of them)

Since test file for TargetResourceHandler was restructured in #101, this PR should only merged after #101 was merged


u413-284-si (approved):
It could be debated whether or not replaceAlias() is a function specific to the TargetResourceHandler and therefore should not be in utilities, but a method of the class. But that is rather of academic nature. It gets the job done. Other than that just a few missing newlines.

html/index.html
```diff
@@ -43,6 +50,9 @@ <h1>A Simple Webpage</h1>
 		<li><a href="./directory/">Directory Listing</a></li>
 		<li><a href="broken_link">Nonexistent page - custom error page for 404</a></li>
 		<li><a href="secret">Links to a directory with missing slash, auto redirect. Also custom error for 403</a></li>
+		<li><a href="/alias/cat.jpg">This link points to /alias/cat.jpg - but there is not /alias directory. It will be
+				replaced and then point to the correct location</a></li>
 	</ul>
 </body>
-</html>
+
+</html>
```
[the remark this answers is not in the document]
gwolf-011235: Added

test/test_utils.cpp
```diff
@@ -15,3 +15,39 @@ TEST(getFileExtension, Extension)
...
+
+TEST(replaceAlias, ReplacementIsEmpty)
+{
+	EXPECT_EQ(webutils::replaceAlias("/path/to/somewhere", "/path", ""), "/to/somewhere");
+}
+
+TEST(replaceAlias, MultipleInstancesOfAlias)
+{
+	EXPECT_EQ(webutils::replaceAlias("/alias/alias/alias/alias/", "/alias", "/replacement"), "/replacement/alias/alias/alias/");
+}
```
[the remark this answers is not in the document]
gwolf-011235: Added

Fun fact: I was confused why this would compile since there is a compiler warning for missing newline. Seems like since C++11 there is no warning for missing newline at end of file anymore:
https://stackoverflow.com/a/72409

inc/ConfigFile.hpp
```diff
@@ -27,6 +27,7 @@ struct Location {
 	std::map<statusCode, std::string> errorPage;
 	bool allowedMethods[MethodCount];
 	std::map<statusCode, std::string> returns;
+	std::pair<bool, std::string> alias;
```
QCHR1581, line 30: I'm not sure why the alias is stored as a ``std::pair<bool, std::string>``.

If I'm right, the bool is used to check if an alias has been set. But can't we just check if the alias string is empty instead? If it's empty, that means no alias was specified.

gwolf-011235: > If I'm right, the bool is used to check if an alias has been set.

Yes you're correct.
 
> But can't we just check if the alias string is empty instead?

I thought if you want to define as alias an empty string. For me it doesn't make much sense but at least nginx accepts it.

But I have to admit I just tested it in the configfile but not what it actually does. This is what I found out:
I defined a location "/hello/" with alias "" [empty string]. On server level there was no root set.
I made a request to "hello/". It added /etc/nginx/ in front, so the actual request was not "" but "/etc/nginx/". 
When I set a server level root it made no difference (which makes sense since these two directives are mutually exclusive).
I could access a file I created at /etc/nginx/index.html with the uri "/hello/index.html" because it was translated to /etc/nginx/index.html

It seems that it has to do with how nginx resolves relative paths: /etc/nginx is the --prefix path of my nginx (checked with nginx -V). If you have a relative path nginx starts at the prefix path. E.g. an alias which does not start with / cannot be an absolute path since it does not start at /. 
If I set as alias "whatever" (no slash) it will replace /hello/ with "whatever" and then append this to "/etc/nginx/". Also alias = "../" would point to /etc/ (since ../ moved one level up).

I think relative path resolution can get messy pretty fast. 
I suggest we only allow absolute paths, which means we would need to check at parsing if alias start with slash "/". In this case we could also drop the std::pair<bool, std::string> like you suggested.

QCHR1581: Wow, I didn't think about the possibility to also specify an empty string as an alias. 

Even though I have to admit that I explicitly check this case for the directive server_name and accept it (nginx says it's okay and I also tested root now. Root also accepts an empty string).

That's an interesting behaviour of nginx. 

I will definitely check if an alias starts with ``/`` and if not I would throw an exception that only absolute paths are excepted. I will also look into the behaviour when an empty string is specified for server_name and root.

gwolf-011235: I think root will show the same behavior: if it's a relative path (so no / at start) it will append it to --prefix path

QCHR1581: Yup, that's probably true. I am going to test it

gwolf-011235: Refactored std::pair into std::string alias.

I will open issue to track that we don't want to implement relative path resolution. Meaning parsing needs to make sure root and alias start with '/'

QCHR1581: Great. Parsing will check, if root and alias start with ``/``.

QCHR1581 (approved):

PR #105, opened 2024-12-12
69 use default config file if no file provided at prog call

gwolf-011235:
### Overview

1. Adds macro DEFAULT_CONFIG_PATH 
2. Adds macros STRINGIZE() and XSTRINGIZE()

#### 1.  Macro DEFAULT_CONFIG_PATH
The value of this macro is used, when no config file path is provided. Macro is used so it can be overwritten at compile time. Has default value of ./config_files/standard_config.conf

Main now accepts zero or one args.

#### 2. Macros STRINGIZE() and XSTRINGIZE()
When an alternative path is provided at compile time via -D one would have to use escaped double quotes: the preprocessor just substitutes the value. This means `-D DEFAULT_CONFIG_PATH=./other/path` would not work > it would replace DEFAULT_CONFIG_PATH with ./other/path, which is not a string literal (since no quotes) and it wouldn't compile. With these two macros the value gets expanded and then stringized. This is also the reason why the default value of DEFAULT_CONFIG_PATH is not put in quotes.


u413-284-si (approved):
What kind of devilry is this

src/main.cpp
```diff
@@ -1,14 +1,41 @@
 #include "ConfigFileParser.hpp"
-#include "ProcessOps.hpp"
 #include "Server.hpp"
 
-int main(int argc, char** argv)
+/**
+ * @brief Stringize the result of expansion of a macro argument
+ *
+ * If the macro DEFAULT_CONFIG_PATH is defined with -D at compile time the literal value would be inserted. Since it has
+ * to be a string one would have to put the vlue in escaped double quotes. This macro in combination with STRINGIZE(s)
```
QCHR1581, line 8: Typo: "value"

QCHR1581 (changes requested):
Works great!

Current default config file causes segfault, but this know and fixed in upcoming PR for parsing. 

Only found a typo.

src/main.cpp
gwolf-011235: Changed!

QCHR1581 (approved):

PR #108, opened 2024-12-13
104 request body not processed when read into connectionm buffer together with request headers

### LINKED ISSUE #104: Request body not processed when read into connection.m_buffer together with request headers
When a complete request header has been identified by a received \r\n\r\n the entire connection.m_buffer is displayed. It shows that the request body is contained in the buffer. However the body is not subsequently processed by the Server, notably the function handleCompleteRequestHeader() seems to swallow the body without forwarding it to connectionReceiveBody().

u413-284-si:
## Overview
1. Add custom form to send POST requests with plain text via Browser (tested with Firefox)
2. Add handleBody() function in Server
3. Fix the handling of body if it remains in the buffer after header is processed

### re 1.
- a simple form to be able to send POST requests with a body in the format text/plain
- serves to prove our ability to handle POST requests  (in the format text/plain) -> further formats may be added later on

### re 2.
 - serves to separate the receiving/reading of data from the socket from the processing of the body data inside of connectionReceiveBody()

### re 3.
- uses new function handleBody() to process remaining data in the connection.buffer after the header is processed (which is equal to the body/part of the body)

closes #104


gwolf-011235 (approved):
Good catch of this bug!

Small thing: Autoformatter was not active, needs to be run over file.
I think the cgi stuff could potentially be another issue?

cgi-bin/time.py
```diff
@@ -2,7 +2,7 @@
 
 import datetime
 
-print("HTTP/1.1 200 OK")
+#print("HTTP/1.1 200 OK")
```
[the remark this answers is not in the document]
u413-284-si: The scripts do not necessarily provide a status, but they [can](https://www.rfc-editor.org/rfc/rfc3875.html#section-6.2.1). If that is the case we should retrieve it and create the status line in our ResponseBuilder accordingly. Will create an issue for it

src/Server.cpp
```diff
@@ -1015,6 +1010,8 @@ void handleCompleteRequestHeader(Server& server, int clientFd, Connection& conne
 	if (connection.m_request.httpStatus == StatusOK && connection.m_request.hasBody) {
 		connection.m_status = Connection::ReceiveBody;
 		connection.m_buffer.erase(0, connection.m_buffer.find("\r\n\r\n") + 4);
+		if (!connection.m_buffer.empty())
+			handleBody(server, clientFd, connection);
```
QCHR1581, line 1014: The test ``POSTRequest`` from the ``HandleCompleteRequestHeaderTest`` test suite fails. 

This is the case because the test checks, if the ``m_connection.m_status`` equals ``Connection::ReceiveBody``, but the function ``handleBody`` sets it to ``Connection::BuildResponse``. 
Which is the result of the passed checks, that the body is complete and the request does not use CGI. 

Therefore this test should be adapted to the new implementation.

QCHR1581 (changes requested):
Tested the upload file page with Brave and it worked great!

Found one test which failed. This is caused by the additional function ``handleBody`` in the function ``handleCompleteRequestHeader``. This new functions sets the status not as expected according to the test.

src/Server.cpp
u413-284-si: Well recognised! The new procedure processes the body immediately without waiting to receive it, does so and then continues to build the response. Test adapted

QCHR1581 (approved):

PR #110, opened 2024-12-15
107 set up integration end2end testing

gwolf-011235:
### Overview

1. Adds integration tests with pytest
2. Adds macro for conditional code removal of DEBUG messages
3. Update coverage gutter search path

### 1. Adds integration tests with pytest

Pytest is an easy to use test framework written in Python [[LINK](https://docs.pytest.org/en/stable/)]. 
In combination with the libraray requests [[LINK](https://requests.readthedocs.io/en/latest/)] we can easliy write tests to check behavior of our server.

#### dev container update
Pytest and requests exist as debian package. These are not the newest versions but they'll do. The packages can be easily added. Dev container needs to be rebuild.

#### test directory layout

A new subdirectory `integration` houses all python tests. Right now there are two "packages" as subdirectories server_response and server_management. This separation was necessary because of the test structure: 
 - server_response uses a single server instance to send requests to 
  - server_management is for starting and stopping server instances (right now only signals test)

The src files for unittest have been moved to subdirectory /unit. This means this PR has many files changed, but they were only moved

#### test_files

I added some example tests. Some are commented out because they don't work right now (e.g. POST, which is fixed via #108 ). Many more can be added. 
Some notes (see docs of pytest for more details):
- Each file is a module as is denoted by test_* (this is what pytest searches for)
- Each test function starts with test_* (again for pytest)
- There are conftest.py files which can configure test environment. Main use is for test fixtures of pytest. Right now we have one fixture which starts a server instance and keeps it running. Since this is defined in the file of server_response package it is not used in server_management

#### Coverage
We can also start kcov in front of the server to capture coverage. In this mode the tests take longer, because of longer startup

#### Makefile
There are new targets:
- make test2: runs integration tests
- make coverage2: runs integration tests with coverage

### 2. Adds macro for conditional code removal of DEBUG messages

Apart from increasing the loglevel, we can now remove the debug messages entirely, instead of just silencing them. If the macro DEBUG_MSG is not defined the macro LOG_DEBUG is replaced with `if (0) std::err`. Since this code does nothing (it is always false) it should be removed by the compiler, increasing our speed (since we don't even send the debug msg).
The macro is activated by default in Makefile, but can be deactivated by setting variable NODEBUG=1 

### 3. Update coverage gutter search path

Coverage gutter search path now searches for `.vscode/coverage/*`. This way it also looks in subdirectory to find cov.xml. Beforehand we created cov.xml in .vscode/coverage/ with running coverage two times. I found out that kcov produces cov.xml output anyway, but stores it in subdirectory. So now coverage gutters work without this extra step. Updated target coverage accordingly.


Makefile
```diff
@@ -40,6 +40,13 @@ else
 	ASAN =
 endif
 
+# Set NODEBUG=1 to compile with debug messages
```
u413-284-si, line 43: You mean "without"

src/EpollWrapper.cpp
```diff
@@ -60,8 +60,9 @@ int EpollWrapper::waitForEvents()
 		throw std::runtime_error("epoll_wait:" + std::string(strerror(errno)));
 	}
 
-	if (nfds == 0)
+	if (nfds == 0) {
```
u413-284-si, line 63: The brackets are superfluous

u413-284-si (approved):
Looking good! Just a couple minor things

Makefile
gwolf-011235: Yes, changed

src/EpollWrapper.cpp
gwolf-011235: We get a "Add explicit braces to avoid dangling else" error without them.

I think it's because of the macro expansion. After the change the code expands to
```
if (nfds == 0)
	if (0) std::cerr << "epoll_wait: Timeout";
else
	if (0) std::cerr << "epoll_wait: " << nfds << " events";
``` 
After the No Ops are removed we have
```
if (nfds == 0)
else
``` 
which has a dangling else. 
With brackets we have
```
if (nfds == 0) {
}
else
``` 
Which is valid. I think the if check will be removed anyway. But it still throws an error

QCHR1581 (approved):
Cool python framework for e2e testing!

Tests work fine besides ``make coverage2`` for me. But this is probably the case, because I use the inferior MacOS compared to Linux ;)

PR #111, opened 2024-12-15
80 implement method delete

### LINKED ISSUE #80: Implement method DELETE
The DELETE method removes the target resource.

More specific: RFC 9110 Section 9.3.5: [DELETE](https://www.rfc-editor.org/rfc/rfc9110.html#name-delete)

Right now there is a conflict with this method and TargetRessourceHandler: if the target resource is a directory and the location doesn't have autoindex, the status code is set to Forbidden (403). This causes an error body to be constructed in ResponseBodyHandler.
The error code should probably be set later in ResponseBuilder.

u413-284-si:
## Overview
1. Add DeleteHandler class
2. Add deleteFile() and deleteDirectory() to FileSystemPolicy
3. Add DirectoryGuard subclass to FileSystemPolicy
4. Integrate MethodDelete into ResponseBodyHandler, Server
5. Fix Content-Type recognition: json
6. Add unittests

### re 1.
- New DeleteHandler to manage DELETE requests
- basically same structure as AutoIndexHandler and FileWriteHandler
- differentiates between several cases: 
1. files
2. directories
3. nonExistent resource (sets status code 404)
4. permission denied (sets status code 403)
5. any other error (sets status code 500)
- returns string containing json body on success, nothing in error case

### re 2.
- deleteFile() uses remove to delete files, throws on error
- deleteDirectory() works recursively to be able to delete non-empty directories, it will traverse the whole directory tree and delete all files and directories in a bottom-to-top approach and finally delete the empty target directory
- in case of errors the throw will be caught in the DeleteHandler

### re 3.
- due to the possibility of a throw while traversing the directory structure, a "guard" for the DIR* pointer was required
- to open a directory a DIR* pointer is created and needs to be closed when done in order to have no leaks
- w/o the guard a throw would leave (potentially multiple) pointers open
- the guard serves to close the pointer upon leaving its scope by utilising its dtor to call closedir
- the guard is implemented as nested subclass in FileSystemPolicy as this is the only class to require it

### re 4.
- MethodDelete is being handled in the ResponseBodyHandler
- check added to allow for directories to be deleted even if they don't have an autoindex, similar to the check allowing for non-exisiting files to be created by POST requests

### re 5.
- the response bodies of POST and DELETE requests are formatted in json
- added check in appendHeaders() to assign Content-Type: application/json in that case

### re 6.
- unittests to test all cases of DeleteHandler
- additional request reqDel to test manually (of in future with pytest)

closes #80


src/FileSystemPolicy.cpp
```diff
@@ -182,3 +183,87 @@ long FileSystemPolicy::getFileSize(const struct stat& fileStat) const
...
+		else {
+			errno = 0;
+			if (unlink(fullPath.c_str()) != 0)
+				throw std::runtime_error("unlink(): " + std::string(strerror(errno)));
+		}
+	}
+
+	// Remove the empty directory itself
+	errno = 0;
+	if (rmdir(path.c_str()) != 0)
```
gwolf-011235, line 241: We're not allowed to use rmdir. But std::remove() also works so also deleteFile

src/ResponseBuilder.cpp
```diff
@@ -102,7 +102,10 @@ void ResponseBuilder::appendHeaders(const HTTPRequest& request)
 {
 	if (!m_responseBody.empty()) {
 	// Content-Type
-		m_responseHeader << "Content-Type: " << getMIMEType(webutils::getFileExtension(request.targetResource)) << "\r\n";
+		if (m_responseBody.find('{') != std::string::npos)
+			m_responseHeader << "Content-Type: application/json\r\n";
```
gwolf-011235, line 106: This makes every file with a curly bracket a json file.

I would change m_request.targetResource to "created.json" or "deleted.json" and add a MIME mapping for .json.
The target resource is not get sent back, so we can reuse it. See for example setDefaultErrorPage() which sets it to "error.html" to get html mapping

inc/FileSystemPolicy.hpp
```diff
@@ -9,6 +9,7 @@
 #include <iostream>
 #include <sys/types.h>
 #include <dirent.h>
+#include <unistd.h>
```
gwolf-011235, line 12: This is new, I think it can be removed? Maybe for forbidden c functions

gwolf-011235 (changes requested):
Looking good!

I think we can consolidate some code re Directory.
Some forbidden functions, but should be easy to change.

inc/DeleteHandler.hpp
```diff
@@ -0,0 +1,25 @@
...
+#include "FileSystemPolicy.hpp"
+#include "Log.hpp"
+#include "utilities.hpp"
+
+#include <sstream>
+#include <stdexcept>
+#include <string>
+
+/**
+ * @brief Class to handle the DELETE requests.
```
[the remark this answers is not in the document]
u413-284-si: Changed

inc/FileSystemPolicy.hpp
```diff
@@ -36,4 +39,17 @@ class FileSystemPolicy {
 	virtual void writeToFile(const std::string& path, const std::string& content) const;
 	virtual std::string getLastModifiedTime(const struct stat& fileStat) const;
 	virtual long getFileSize(const struct stat& fileStat) const;
+	virtual void deleteFile(const std::string& path) const;
+	virtual void deleteDirectory(const std::string& path) const;
```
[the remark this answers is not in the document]
u413-284-si: Moved deleteDirectory() to DeleteHandler

inc/FileSystemPolicy.hpp
```diff
@@ -36,4 +39,17 @@ class FileSystemPolicy {
...
+
+private:
+    class DirectoryGuard {
```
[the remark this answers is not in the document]
u413-284-si: Thanks for making me aware of this, certainly it makes sense to use this class then. Consolidated

src/FileSystemPolicy.cpp
```diff
@@ -182,3 +183,87 @@ long FileSystemPolicy::getFileSize(const struct stat& fileStat) const
...
+ * If the file cannot be deleted, it throws a runtime error with the
+ * appropriate error message.
+ *
+ * @param path The path to the file to be deleted.
+ * @throws std::runtime_error if the file cannot be deleted.
+ */
+void FileSystemPolicy::deleteFile(const std::string& path) const
+{
+	errno = 0;
+	if (remove(path.c_str()) != 0)
```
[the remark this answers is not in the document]
u413-284-si: CHanged

src/FileSystemPolicy.cpp
```diff
@@ -182,3 +183,87 @@ long FileSystemPolicy::getFileSize(const struct stat& fileStat) const
...
+
+		std::string fullPath = std::string(path) + "/" + entry->d_name;
+
+		struct stat info = getFileStat(fullPath);
+		// NOLINTNEXTLINE: misinterpretation by HIC++ standard
+		if (S_ISDIR(info.st_mode))
+			deleteDirectory(fullPath);
+		else {
+			errno = 0;
+			if (unlink(fullPath.c_str()) != 0)
```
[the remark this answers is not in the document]
u413-284-si: Changed

src/FileSystemPolicy.cpp
```diff
@@ -182,3 +183,87 @@ long FileSystemPolicy::getFileSize(const struct stat& fileStat) const
...
+ * @param path The path to the directory to be deleted.
+ * @throws std::runtime_error if any file or directory cannot be deleted.
+ */
+// NOLINTNEXTLINE (misc-no-recursion): recursion is being handled
+void FileSystemPolicy::deleteDirectory(const std::string& path) const
+{
+	DirectoryGuard dir(openDirectory(path));
+
+	struct dirent* entry = NULL;
+	while ((entry = readdir(dir.getDir())) != NULL) {
```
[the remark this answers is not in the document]
u413-284-si: Changed

src/FileSystemPolicy.cpp
u413-284-si: Changed

u413-284-si: Ah no, actually std::remove only deletes directories starting with C++17, so we don't rly have another choice but to use rmdir. Any suggestions?

inc/FileSystemPolicy.hpp
u413-284-si: Removed

src/ResponseBuilder.cpp
u413-284-si: This would also not be correct as this would require/ infer that all POST requests have a json body which is not necessarily the case. The json content-type only refers to the response body which is in the json format. The target resource can have any format.

It would be necessary to find a distinctive feature in the json format in order to recognise it. As we only have responses in either HTML or json, I chose the bracket. Alternatively one could always use a content-type json whenever a successful POST or DELETE response is sent.

src/FileSystemPolicy.cpp
gwolf-011235: In that case I think we simply can't delete a directory? I didn't find any alternative, and since `rmdir` is not allowed function we simply can't / are allowed to do it?

We could check if it's a regular file and only delete in that case. Else return error 400 or maybe 403?
Or we ask for clarification in discord, but I think the answer will be: "look at list".

src/ResponseBuilder.cpp
gwolf-011235: This adds the header for _our response_. 

What came in as POST is already processed, the content-type was given to us by the browser. We send our response, which in case of POST and DELETE right now is always in .json (from FileWriteHandler, DeleteHandler) or in case of GET can be .html, .txt, .png, whatever.

The Bracket '{' or byte 0x7B is far from distinct. Our index.html has brackets for styling. If you start server and navigate to index.html, the browser will give you a .json representation and will not render html (since we say its json). I think the possibility of a binary format like png having this byte is also high. 
MIME mapping is very "dumb" in that in only cares for file extension and not magic numbers or stuff.
Since the target resource for POST and DELETE is not used anymore at this stage (we already created / deleted the file) we can just reuse it to identify our response as json

src/FileSystemPolicy.cpp
u413-284-si: I agree that asking on Discord would be fruitless. I can just set an error in the switch case, but it is rly a shame to leave it out just because of such senseless restrictions. It works wonderfully as such. In that case the bug about removing the dots in the target resource would not be relevant anymore, no?

src/ResponseBuilder.cpp
u413-284-si: To my understanding, what you're asking to do is to set the target resource in the request to:

E.g. /uploads/myfile.json

POST /uploads/myfile.json HTTP/1.1
...

Then define a mime_mapping element which recognises the .json extension and then set the Content-Type header in the response to application/json.

But this would imply that the posted content is in json format, although it can be anything.

If you mean sth else, please illustrate, best with an example.

src/FileSystemPolicy.cpp
gwolf-011235: If the product owner say no we have to be obedient ;)

No dot removal is still necessary since you could still GET a sensitive file of the filesystem and hack the server

src/ResponseBuilder.cpp
gwolf-011235: No. After successfully processing the DELETE request, reuse our variable targetResource and set it to "deleted.json". Like so:
```
if (m_request.method == MethodDelete) {
  DeleteHandler deleteHandler(m_fileSystemPolicy);
  m_responseBody = deleteHandler.execute(m_request.targetResource, m_request.httpStatus);
  if (m_responseBody.empty())
    handleErrorBody();
  m_request.targetResource = "deleted.json"; <--- here we reuse variable
}
```

Example: existing implementation of ResponseBodyHandler::setDefaultErrorPage():
```
void ResponseBodyHandler::setDefaultErrorPage()
{
	m_responseBody = getDefaultErrorPage(m_request.httpStatus);
	m_request.targetResource = "error.html";
}
```

When we return to ResponseBuilder, the target resource has a json extension. If you add a mapping for it, then the response would have the correct content-type: json

src/FileSystemPolicy.cpp
u413-284-si: Oh well, will set the error, probably Forbidden. And comment out the deleteDirectory(). But once the eval is done I'm activating it again!

src/ResponseBuilder.cpp
u413-284-si: Achsooo, ja that can be done

u413-284-si: Implemented

src/FileSystemPolicy.cpp
u413-284-si: Done

src/DeleteHandler.cpp
```diff
@@ -0,0 +1,108 @@
...
+// {
+// 	Directory dir(m_fileSystemPolicy, path);
+
+// 	struct dirent* entry = NULL;
+// 	while ((entry = m_fileSystemPolicy.readDirectory(dir.getDir())) != NULL) {
+// 		// Skip `.` and `..` entries
+// 		if (strcmp(entry->d_name, ".") == 0 || strcmp(entry->d_name, "..") == 0)
+// 			continue;
+
+// 		std::string fullPath = std::string(path) + "/" + entry->d_name;
```
gwolf-011235, line 94: Moot since not allowed, but with existing function of Directory can be written as 
```
const std::vector<std::string> entries = dir.getEntries();
for (std::vector<std::string>::const_iterator iter = entries.begin(); iter != entries.end(); ++iter) {
  // Skip `.` and `..` entries
  if ((*iter == ".") || (*iter == ".."))
    continue;

  std::string fullPath = path + "/" + *iter;

  struct stat info = m_fileSystemPolicy.getFileStat(fullPath);
  // NOLINTNEXTLINE: misinterpretation by HIC++ standard
  if (S_ISDIR(info.st_mode))
    deleteDirectory(fullPath);
  else
    m_fileSystemPolicy.deleteFile(fullPath);
}
```

test/test_DeleteHandler.cpp
```diff
@@ -0,0 +1,100 @@
...
+	// Act
+	std::string responseBody = m_deleteHandler.execute(m_path, m_statusCode);
+
+	// Assert
+	EXPECT_EQ(responseBody,
+		"{\n\"message\": \"File deleted successfully\",\n\"file\": \"" + m_path
+			+ "\"\n}\n");
+}
+
+TEST_F(DeleteHandlerTest, DeleteDirectory)
```
gwolf-011235, line 44: Need to comment out since we cant use deleteDirectory()

gwolf-011235 (approved):
Looks good!

Some comments re (sadly not usable) deleteDirectory, but not really needed anymore (except unit test)

inc/DeleteHandler.hpp
```diff
@@ -0,0 +1,27 @@
...
+
+public:
+	explicit DeleteHandler(const FileSystemPolicy& fileSystemPolicy);
+	std::string execute(const std::string& path, statusCode& httpStatus);
+
+private:
+	const FileSystemPolicy& m_fileSystemPolicy;
+	std::stringstream m_response;
+
+	void deleteDirectory(const std::string& path) const;
```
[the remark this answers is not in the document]
u413-284-si: Comment introduced

inc/Directory.hpp
```diff
@@ -17,6 +17,7 @@ class Directory {
 	~Directory();
 
 	std::vector<std::string> getEntries();
+	DIR* getDir() const;
```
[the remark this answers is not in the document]
u413-284-si: True, changed

src/DeleteHandler.cpp
```diff
@@ -0,0 +1,108 @@
...
+			// 		   << "\"directory\": \"" << path << "\"\n"
+			// 		   << "}\n";
+			break;
+
+		case FileSystemPolicy::FileNotExist:
+			httpStatus = StatusNotFound;
+			break;
+
+		case FileSystemPolicy::FileOther:
+			httpStatus = StatusInternalServerError;
```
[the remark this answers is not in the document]
u413-284-si: I saw that you already changed that in one of your PRs. Changed

src/DeleteHandler.cpp
```diff
@@ -0,0 +1,108 @@
...
+			break;
+		}
+	} catch (const std::runtime_error& e) {
```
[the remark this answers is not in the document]
u413-284-si: Added doxygen

src/DeleteHandler.cpp
u413-284-si: Changed

u413-284-si: The rmdir can be added to FileSystemPolicy after eval

test/test_DeleteHandler.cpp
u413-284-si: Comment introduced

QCHR1581 (approved):
Works perfectly fine. 

Unfortunate that we are not allowed to use rmdir to also delete a directory properly.

PR #116, opened 2024-12-17
106 implement maxbodysize

gwolf-011235:
### Overview

1. implements maxBodySize from Location to check if we still want receive bytes.
2. fixed a little bug: we would only receive upto Server::s_clientBodyBufferSize beforehand.
3. little refactor of Location and ConfigServer, updating Ostreaminserter
4. added unittests for connectionReceiveBody
5. activated and added integration tests for POST --> found a bug (?)

### 1. implements maxBodySize 

- We just used static Server::s_clientMaxBodySize in connectionReceiveBody
- Removed static variable and replaced with maxBodySize of active location for connection
- connectionReceiveHeader and connectionReceiveBody now check if respective value is >= not just == buffer size. Logically we should never be bigger (since we check beforehand). But it makes testing easier.

### 2. Bugfix

- We only checked after receiving, if we had received Server::s_clientMax**Body**Size already.
- But how many bytes we would read into buffer was (only) dependent on how much bytes we already received. So we could read upto s_clientBody**Buffer**Size
- Now we check if bytes available (= maxBodySize - received bytes) is bigger than our recv buffer size.
  - if yes: we can read our whole buffer size
  - if no: we can only read buffer size - bytes available

### 3. Refactor

- Removed empty vector inits from constructor of Location and ConfigServer
- Renamed isAutoindex to hasAutoindex
- Moved ostream definition for method to method.hpp
- Updated ostream inserter for location to print all values
- introduced new constant g_oneMegabyte

### 4. Unittests for connectionReceiveBody

- similar tests like connectionReceiveHeader

### 5. integration tests for POST 

- activated existing tests and adapted them slightly
- commented out debug printing of received body in connectionReceiveBody > was hard to read
- also added new tests for big file (so requestHeaderBuffer cannot have all of it) and too big file (will be rejected)
- BUG(?): if body is sent with chunked encoding, and it contains a newline \n, it will give an error ERR_MISS_CRLF in RequestParser. I think because getline eats the newline
Reproduce: add a \n in function generate_chunks() of test_POST.py


u413-284-si (approved):
Looks good. Should have a quick sesh about the mentioned potential bug

QCHR1581 (approved):
Good job, lgtm.

PR #117, opened 2024-12-20
114 requested uri dot removal

gwolf-011235:
Implements dot removal algorithm as described by RFC 3986 Sect 5.2.4 [[LINK](https://datatracker.ietf.org/doc/html/rfc3986#section-5.2.4)]

As RFC 3986 is general specification for all URIs not the whole algorithm was implemented to keep it simpler. Since we know a request path has to be an absolute reference (needs to start with '/') we can ignore parts that deal with relative parts.

Implemented in RequestParser(), as static function.

Added unittests and integration tests. Note: integration test needed to be implemented at lower level, as request library automatically removes dot segments.


src/RequestParser.cpp
```diff
@@ -1034,3 +1037,99 @@ bool RequestParser::isValidHostname(const std::string& hostname)
...
+/**
+ * @brief Removes last segment of a path denoted by '/'.
+ *
+ * @param output Reference to buffer containing path.
+ */
+static void removeLastSegment(std::string& output)
+{
+	if (!output.empty()) {
+		size_t lastSlash = output.find_last_of('/');
+		if (lastSlash != std::string::npos) {
```
u413-284-si, line 1083: Could drop some brackets

src/RequestParser.cpp
```diff
@@ -1034,3 +1037,99 @@ bool RequestParser::isValidHostname(const std::string& hostname)
...
+ * @param path Where to remove possible dot segments from.
+ * @return std::string String with dot segments removed.
+ */
+std::string RequestParser::removeDotSegments(const std::string& path)
+{
+	std::string output;
+	std::string::const_iterator iter = path.begin();
+
+	while (iter != path.end()) {
+		if (*iter++ == '/') {
```
u413-284-si, line 1117: Would remove the post increment for better readability. Just add a iter++ in the line after. Then the output can be += *iter.

u413-284-si (changes requested):
Works well, small adjustments requested

src/RequestParser.cpp
gwolf-011235: Removed brackets

src/RequestParser.cpp
gwolf-011235: Removed post increment, added iter++ in if and else clause

u413-284-si (approved):

QCHR1581 (approved):
Successfully prevents directory traversal!

I don't think we have to reproduce the behaviour of nginx for such cases, because this is nowhere defined in the subject, but I think it's still nice to know:

Nginx produces for all test requests you implemented a bad request. 
This is the case, if there is an attempt to access data outside of the root directory. 

It does not produce a bad request, if you only try to access data inside the root directory.

Example (using telnet): 
```bash
GET /uploads/.. HTTP/1.1
Host: localhost:8080
```

This results in this request path: ``\``
For my config then nginx returned the index file and the status code 200.

gwolf-011235:
> Nginx produces for all test requests you implemented a bad request.
This is the case, if there is an attempt to access data outside of the root directory.

Good point!

So as soon as we try to remove a non existing segment, we could return bad request: E.g. "/search/../.././hello/":
1. "/search" > append /search
2. "/.." > remove one segment (/search)
3. "/.." > remove one segment > no segment left, bad request

In that case we could also remove the check for output.empty() since it always has to have something in there: RequestParser ensures that each request starts with '/'. It can only be empty after a "/.." which would already return BadRequest

gwolf-011235:
Adapted to throw BadRequest if request line travels outside of root (e.g. "/.." or "/search/../..") 

We still need to check if buffer is empty at the end: If it does not travel outside but would only leave slash (like the given example "/uploads/.."). The double dot removes the last segment leaving us with empty buffer, but it's still valid. So at the end we return a single slash.

QCHR1581:
Great that you adapted the implementation according to the behaviour of Nginx!!

Issue #127, opened 2025-01-02
Bug in cleanup of closed connections


gwolf-011235:
When we close a connection we close fd and set status to Connection::Closed. After every epoll trigger we loop through connections with cleanupClosedConnections() and remove those with status Connection::Closed. I implemented delayed removal for efficency: we don't remove every closed one individually, but remove all closed ones in a loop.

However this leads to a bug: 
- epoll reports fd 8 (connection event) and fd 4 (server event).
- processing of fd 8 leads to connection close. The fd is closed and status is set to Connection::Closed. It is not removed yet from ConnectionsMap.
- processing of fd 4 leads to accept of new connection. It receives the next unused fd number: 8
- we try to add the new connection to ConnectionMap but receive error: slot 8 is already taken - by the closed connection.

Suggestion:

- [x] implement similar function to existing Server::removeCGIFileDescriptor(): removeEvent, close fd and remove connection from map in one function.
- [x] Remove function cleanupClosedConnections()

Alternative: 
- Only set status to Connection::Closed, leave fd open, and close it in cleanupClosedConnetion(). Would keep structure like it is now. But I think there should be no performance loss by removing it immediatly. More likely is a perofrmance gain, since we don't loop through all connections everytime.
- We could also put all connections in a std::vector and instead of deleting them, clean and reuse them. Would be a bigger refactor, therefore I decided against it (for now :D ).

PR #129, opened 2025-01-03
127 bug in cleanup of closed connections

gwolf-011235:
Fixes bug as described in #127.

Also added integration test for this case, which now passes.


u413-284-si (approved):
LGTM, nothin to be added

Issue #133, opened 2025-01-10
FileWriteHandler does not set StatusForbidden


gwolf-011235:
FileWriteHandler and DeleteHandler have a similiar structure. However DeleteHandler catches NoPermissionException if FileSystemOps throws it. It then sets passed httpStatus to StatusForbidden.

FileWriteHandler catches only std::runtime_error. It will catch NoPermissionException (since it derives from runtime_error) but only sets StatusInternalServerError.

Suggestion - make it more similar to DeleteHandler
- pass `statusCode& httpStatus` to FileWriteHandler
- update status inside of FileWriteHandler::execute() : Created, NoPermission, InternalServerError
- extra catch block for NoPermissionException 
- outside of FileWriteHandler would need to check if statusCreated -> add header; and like DeleteHandler check if response.empty() -> handleErrorBody()

PR #134, opened 2025-01-11
133 filewritehandler does not set statusforbidden

gwolf-011235:
Fixes bug as described in #133.

FileWriteHandler and DeleteHandler now behave more similar: they get passed statusCode and set it themselves. FileWriteHandler also catches NoPermissionException.

Added unit and integration test cases

### DIFF
--- a/src/FileWriteHandler.cpp
+++ b/src/FileWriteHandler.cpp
@@ -55,6 +55,10 @@ std::string FileWriteHandler::execute(const std::string& path, const std::string
 	} catch (FileSystemPolicy::NoPermissionException& e) {
 		LOG_ERROR << e.what();
 		httpStatus = StatusForbidden;
+	} catch (FileSystemPolicy::FileNotFoundException& e) {
+		LOG_ERROR << e.what();
+		LOG_ERROR << "Missing directory in path " << path;
+		httpStatus = StatusNotFound;
 	} catch (std::runtime_error& e) {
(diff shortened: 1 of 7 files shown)


src/FileWriteHandler.cpp
```diff
@@ -50,10 +52,13 @@ std::string FileWriteHandler::execute(const std::string& path, const std::string
 					   << "\"status\": \"created\"\n"
 					   << "}\n";
 		}
-		return m_response.str();
+	} catch (FileSystemPolicy::NoPermissionException& e) {
```
QCHR1581, line 55: P1: The function ``writeToFile`` can throw those exceptions:

- FileNotFoundException
- NoPermissionException
- std::runtime_error

You catch NoPermissionException and std::runtime_error. But FileNotFoundException is missing.
A catch for FileNotFoundException would make sure we handle this exception properly.

QCHR1581 (changes requested):
Tests work perfectly. 

Saw that a catch for one exception should be implemented, in the case it gets thrown.

src/FileWriteHandler.cpp
gwolf-011235: True that!
We would still catch it since it's derived from std::runtime_error. But we can return a more explicit error message.

Added catch, error_msg, unit and integration test

QCHR1581 (approved):

Issue #126, opened 2025-01-02
Status 405 response should include allow header


gwolf-011235:
From [RFC 9110 Sect 10.2.1 Allow](https://www.rfc-editor.org/rfc/rfc9110.html#name-allow): An origin server **MUST** generate an Allow header field in a [405 (Method Not Allowed)](https://www.rfc-editor.org/rfc/rfc9110.html#status.405) response and MAY do so in any other response. An _empty Allow field value_ indicates that the resource allows no methods, which might occur in a 405 response if the resource has been temporarily disabled by configuration.

Suggestion:

- [ ] Add a simple check in responsebuilder: (if status == 405) add header Allow. Loop through allowed methods of location and add to header

PR #140, opened 2025-01-17
126 status 405 response should include allow header

gwolf-011235:
Adds function to construct Allow header (see also #126).

Added unittest and integration test.


QCHR1581 (approved):
Allow header gets added accordingly to RFC 9110. 

Implemented tests work fine and also manual tests by myself worked. 
Good job!

PR #144, opened 2025-01-19
138 update static webiste

gwolf-011235:
### Added links to all cgi scripts in run_cgi.html. 
- They can be executed via browser
- Modified CGI upperCase.sh 
  - Now works with GET and POST
  - GET uses querystring and POST uses content
- Modified CGI upload.py
  - Beforehand expected querystring and content which is only possible via JavaScript.
  - Now works with application/x-www-form-urlencoded html form --> only creates textfiles. But was kinda expected to do beforehand, since default file name was uploaded_file.txt. Now explicit.
  - Renamed from upload.py to create_textfile.py
- Modified CGI badBoi.sh to include Content-type header 

### Add delete.html to issue DELETE requests 
  - Since browser can't naturally do this JavaScript Magic is used
  - File list of upload directory is created (uses autoindex from our server). Radio button is added to select file. Delete button sends a fetch with DELETE method.

### Fix: Autoindexhandler 
- also receives uriPath to construct better relative links. Beforhand just had href with file name.


QCHR1581 (approved):
Showcases well that we can handle cgi scripts and are able to delete files!

At the moment we only handle one cgi extension at once. 
Therefore you provided another port number when a python script should be executed. 

We could implement displaying scripts dynamically according to a provided extension (by using a drop down menu for example). This is not really necessary, but just an idea from my side. Therefore my approvement for this PR.

html/run_cgi.html
```diff
@@ -14,37 +14,79 @@ <h1>CGI Test</h1>
...
 		<section class="form-section">
-			<h2>Script Upload</h2>
-			<p>Select a file to upload to the server.</p>
-			<!-- Form for file upload -->
-			<form action="cgi-bin/upload.cgi" method="POST" enctype="multipart/form-data">
-				<div class="form-group">
-					<input type="file" id="file" name="file" required>
-				</div>
+			<h2>Script badBoi.sh</h2>
+			<p>This script always returns a 400 error</p>
```
u413-284-si, line 50: Actually the idea behind badBoi.sh is to showcase that a CGI script can return 

1. A random reason phrase
2. Bad Request and this would still be an acceptable Status

u413-284-si (changes requested):
Nice job cleaning up the CGI page!

One thing: the badBoi.sh should showcase that a Status 400 returned by the script is also a valid status being displayed in the response. This isn't the case and leads to the connection being disrupted.

html/run_cgi.html
gwolf-011235: >  the badBoi.sh should showcase that a Status 400 returned by the script is also a valid status being displayed in the response. This isn't the case and leads to the connection being disrupted.

I don't understand. In what way is the connection disrupted? Right now it returns a 400 with custom reason phrase for me.

Or do you just mean to update the description?
"This script returns a 400 error with a custom reason phrase"

u413-284-si:
Also two integration tests fail:

image

html/run_cgi.html
gwolf-011235: I tested with Chrome.

However I can't reproduce on Firefox. It shows page:

image

Also console shows status plus reason phrase:

image


Maybe try to clear cache?

u413-284-si: No I mean, you could have had it working before, then introduced a breaking change, but your cache was still delivering the good website.

But it works on the campus PC?

Weirdly nothing is even being sent:
image

gwolf-011235: Is the server running?

u413-284-si: Yes haha, I am accessing the link via the CGI page

u413-284-si (approved):

Issue #136, opened 2025-01-13
Empty server/location block should not cause error


QCHR1581:
The configs below cause currently an error message:

### Empty server block

```nginx
server{
}
````
### Empty location block

```nginx
server{
           location /path{
           }
}
````

Nginx accepts those configs and therefore those file should not cause any issues for our webserver.

PR #146, opened 2025-01-20
136 empty serverlocation block should not cause error

QCHR1581:
Refer to issue #136 for more information.

# Additional checks

At first the ``serverBlockContent`` and the ``locationBlockContent`` were only checked for a missing semicolon. 

Because there are no semicolons within an empty server block or location block, this check resulted in the error:
```bash
 [ERROR] Unexpected '}'
```

Therefore I added an additional check to ensure that the content is not only composed of whitespace characters.

# Additional function

The first line of the ``locationBlockContent`` contains the location directive, path, and opening curly bracket. 
To enable checking for an empty content, I implemented the ``removeLocationPathLineFromContent`` function. 
This function removes the first line from the locationBlockContent, stores it in a separate string, and returns the resulting string.

With this new string, it is now possible to verify whether the content is empty or consists only of whitespace characters.


src/ConfigFileParser.cpp
```diff
@@ -632,3 +635,21 @@ void ConfigFileParser::skipLocationBlockPath(size_t& index)
...
+
+/**
+ * @brief Removes the location path line from the location block content
+ *
+ * Removes the line which contains the directive 'location' and the corresponding path from the location block content
+ *
+ * @param locationBlockContent The content of the location block
+ * @return std::string The content without the location path line
+ */
+std::string ConfigFileParser::removeLocationPathLineFromContent(const std::string& locationBlockContent) const
```
gwolf-011235, line 647: P3: This creates a new substring of the location content just so you don't have to iterate over the LocationPathLine.

I think you could just use the found index and check from there if nothing is inside:
```
const size_t openingCurlyBracketIndex = locationBlockContent.find('{');
const size_t firstNonWhitespaceAfterCurly = locationBlockContent.find_first_not_of(s_whitespace, openingCurlyBracketIndex + 1);
return (firstNonWhitespaceAfterCurly != std::string::npos);
```

I'd rename to isEmptyLocationBlock()

src/ConfigFileParser.cpp
```diff
@@ -257,7 +257,8 @@ void ConfigFileParser::processServerContent(const ServerBlockConfig& serverBlock
 	ConfigServer server;
 	m_configFile.servers.push_back(server);
 
-	if (isSemicolonMissing(serverBlockConfig.serverBlockContent))
+	if (isSemicolonMissing(serverBlockConfig.serverBlockContent)
+		&& serverBlockConfig.serverBlockContent.find_first_not_of(s_whitespace) != std::string::npos)
```
gwolf-011235, line 261: P3: In combination with my suggestion down below: would pack this into a function isEmptyServerBlock().
I know it's just a line but maybe a tad more readable

test/unit/test_ConfigFileParser.cpp
```diff
@@ -289,6 +291,18 @@ TEST_F(InvalidConfigFileTests, FileMissesServerBlock)
 	EXPECT_EQ(0, virtualServers.size());
 }
 
+TEST_F(ValidConfigFileTests, FileContainsEmptyServerBlock)
+{
+	ConfigFile configFile;
+	EXPECT_NO_THROW(configFile = m_configFileParser.parseConfigFile("config_files/empty_server.conf"));
```
gwolf-011235, line 297: P2 / P3:
Just noticed: the unittest only check if the parser never throws, but they actually never validate what was parsed.
Maybe add one or two tests which validate what content was returned:
- one with dummyConfig (has many values)
- one with empty Config (should all have standard values)

gwolf-011235 (approved):
Nice little change!

Just have some P3 comments.

src/ConfigFileParser.cpp
QCHR1581: Nice approach, thanks! 

Is implemented. 

Little side note:
This line
```cpp
return (firstNonWhitespaceAfterCurly != std::string::npos);
```

Should look like this
```cpp
return (firstNonWhitespaceAfterCurly == std::string::npos);
```

If a non whitespace character can't be found after the curly bracket, the location is empty.

src/ConfigFileParser.cpp
QCHR1581: True, this makes it also more coherent. Is implemented

test/unit/test_ConfigFileParser.cpp
gwolf-011235: Nice

PR #148, opened 2025-01-22
fix: remove check against content length 0

### LINKED ISSUE #139: Content-Length of 0 should be accepted
From [RFC 9110 Sect 8.6. Content-Length](https://www.rfc-editor.org/rfc/rfc9110.html#name-content-length):
- "Any Content-Length field value greater than or **equal to zero** is valid." [emphasis mine]
- A little bit higher: "For example, a user agent normally sends Content-Length in a POST request even when the value is 0 (indicating empty content)."

A POST request with content-length 0 is valid. It could be used to trigger an event on the server which does not need a body (e.g. some api which does something to the server state)

In RequestParser::validateContentLength(...) if value of content-length is 0 we throw with "Invalid HTTP request: Invalid content-length provided" and return a 400 bad request.

I think if we just delete that check, it could still run through and create an empty file?

I noticed it while testing DELETE method with python request > request.delete() sends a content-length: 0, which gets rejected as bad request.

However sending content-length with DELETE is not forbidden. Even a body could be send, although what should happen then is not explicitly defined. 
See [RFC 9110 Sect 9.3.5. DELETE](https://www.rfc-editor.org/rfc/rfc9110.html#section-9.3.5): 
"Although request message framing is independent of the method used, content received in a DELETE request has no generally defined semantics, cannot alter the meaning or target of the request, and might lead some implementations to reject the request and close the connection because of its potential as a request smuggling attack ([Section 11.2](https://www.rfc-editor.org/rfc/rfc9112#section-11.2) of [[HTTP/1.1](https://www.rfc-editor.org/rfc/rfc9110.html#HTTP11)]). A client SHOULD NOT generate content in a DELETE request unless it is made directly to an origin server that has previously indicated, in or out of band, that such a request has a purpose and will be adequately supported.content-length is explicitly"

u413-284-si:
Removes check for content length 0

Added unit test to verify

closes #139

### DIFF
--- a/src/RequestParser.cpp
+++ b/src/RequestParser.cpp
@@ -598,7 +598,7 @@ void RequestParser::validateContentLength(const std::string& headerName, std::st
 		for (size_t i = 0; i < strValues.size(); i++) {
 			char* endptr = NULL;
 			const long contentLength = std::strtol(strValues[i].c_str(), &endptr, constants::g_decimalBase);
-			if ((contentLength == 0) || *endptr != '\0') {
+			if (*endptr != '\0') {
 				request.httpStatus = StatusBadRequest;
 				request.shallCloseConnection = true;
 				throw std::runtime_error(ERR_INVALID_CONTENT_LENGTH);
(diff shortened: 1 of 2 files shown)


gwolf-011235 (approved):
LGTM

Issue #145, opened 2025-01-20
Autoindex is handled instead of POST


gwolf-011235:
In ResponseBodyHandler::execute we check how to construct a body depending on set variables. Right now autoindex is checked before POST request. This means we can't upload to a directory if autoindex is on since we would always generate a autoindex body instead of handling POST request.

Solution: Restructure execute function to better differentiate.

PR #150, opened 2025-01-22
141 cgi handler sets some env vars wrong

### LINKED ISSUE #141: CGI Handler sets some env vars wrong
Some env vars are constructed using "connection.location->root" which does not work if an alias is used.
- PATH_TRANSLATED
- SCRIPT_FILENAME
- SYSTEM_ROOT

I think we could use request.targetResource, however we need to strip it from PATH_INFO.

From [RFC 3875 Sect 4.1.6 PATH_TRANSLATED](https://www.rfc-editor.org/rfc/rfc3875#section-4.1.6): "The PATH_TRANSLATED variable is derived by taking the PATH_INFO value, parsing it as a local URI in its own right, and performing any virtual-to-physical translation appropriate to map it onto the server's document repository structure." 

I think this means we can't just take the root / alias of the found location for the CGI script. But we have to make an extra search for the PATH_INFO value (which could end up in another location) to get PATH_TRANSLATED.

u413-284-si:
Use target resource to determine root path of ENV vars which indicate absolute file system paths to requested resources.
Add function to extract path before script name.

Adapt unit tests.

closes #141


gwolf-011235:
Tl;dr: P1: we need TargetResourceHandler to give us correct PATH_TRANSLATED

From #145: 

"
From [RFC 3875 Sect 4.1.6 PATH_TRANSLATED](https://www.rfc-editor.org/rfc/rfc3875#section-4.1.6): "The PATH_TRANSLATED variable is derived by taking the PATH_INFO value, parsing it as a local URI in its own right, and performing any virtual-to-physical translation appropriate to map it onto the server's document repository structure."

I think this means **we can't just take the root / alias of the found location of the CGI script**. But we have to make an extra search for the PATH_INFO value (which could end up in another location) to get PATH_TRANSLATED.
"

Example:
openAndDisplayFile.sh - A imaginary script that opens a file, read its contents and sends them to server.
We have a `/documents/`- location with root `/html/documents/`

We retrieve /cgi-bin/openAndDisplayFile.sh/documents/myfile.txt

The file path is correctly stored in PATH_INFO. However since PATH_INFO does not say where the file exactly is (it is not translated to the filesystem yet), we want to take PATH_TRANSLATED instead. 
Right now it ends up at /cgi-bin/documents/myfile.txt -> we would need to store our files in the cgi-bin directory.

But specification say "parses the file as a local URI in its own right" - I think we should try to map path with TargetResourceHandler and see where it ends up. So the actual value should be /html/documents/myfile.txt - which makes more sense to me, as we don't need to store everything in cgi-bin (or wherever the script is located)

We do have access to Connection object, so we could construct a TargetResourceHandler and execute it with new "path". However this could change active location of the connection and alter other things in Request object (isDirectory, Status).

I could rewrite TargetResourceHandler::execute so that it doesn't need a Connection but just HTTPRequest, Location and Servers. Then we could pass a dummy value and not the real connection. 
Or we construct a dummy connection object and pass that.

Or did I just misunderstand PATH_TRANSLATED?

gwolf-011235:
Adapted to make path search with TargetResourceHandler

u413-284-si:
I can't pass a review as I originally posted the PR. @gwolf-011235  you have to review yourself so to say. But I had a look at it, thumbs up from me.

gwolf-011235 (approved):
LGTM

PR #151, opened 2025-01-24
145 autoindex is handled instead of post

gwolf-011235:
- Fixed bug: only construct autoindex in case of GET request 
- Refactor ResponseBodyHandler::execute by splitting in up in smaller functions


u413-284-si (approved):
LGTM, nice clean up

Issue #137, opened 2025-01-13
Location path: invalid amount of parameters


QCHR1581:
The configs below cause currently NOT an error message:

### Location contains no path
```nginx
server {
           location {
               root /var/www;
            }
}
```

### Location contains more than one path

```nginx
server {
           location /path /anotherone {
               root /var/www;
            }
}
```

Nginx does not accept both config files. 
Therefore our webserver should also display an error message for such cases.

PR #152, opened 2025-01-24
137 location path invalid amount of parameters

QCHR1581:
Refer to issue #137  for more information.

# Additional checks

Implemented additional check in the function ``isValidBlockBeginn``:

```cpp
if (block == LocationBlock) {
	if (m_configFileContent[index] == '{')
		return false;
	skipLocationBlockPath(index);
}
```
Normally the directive ``location`` gets skipped and after that the path. In the case that no path is provided,  the index gets immediately skipped to the char ``{``. This indicates a lack of a location path. 

Therefore the location block begin is not valid and ``false`` is returned. 

# Refactor

The function ``readServerBlock`` was refactored for the occasion, that a ``location`` directive is reached and the beginn of this block is not valid:

```cpp
while (m_configFileContent[m_configFileIndex] != '}') {
	if (isKeyword(convertBlockToString(LocationBlock), m_configFileIndex)) {
		if (!isValidBlockBeginn(LocationBlock))
			throw std::runtime_error(ERR_LOCATION_INVALID_BEGIN);
		serverBlockConfig.serverBlockContent
			+= m_configFileContent.substr(startIndex, m_configFileIndex - startIndex);
		readLocationBlock(serverBlockConfig);
		startIndex = m_configFileIndex;
	}
	m_configFileIndex++;
}
```
In both cases, when a path is not provided or multiple paths are provided, an error is thrown which indicates an invalid location block begin.


config_files/location_empty_path.conf
```diff
@@ -0,0 +1,12 @@
...
+        listen 127.0.0.1:80;
+        server_name example.com;
+        root /var/www/html;
+
+        location {
+            client_max_body_size 1g;
+            index index.html default.html;
+        }
+	}
+}
```
gwolf-011235, line 12: P3: missing newline

config_files/location_multiple_paths.conf
```diff
@@ -0,0 +1,12 @@
...
+        listen 127.0.0.1:80;
+        server_name example.com;
+        root /var/www/html;
+
+        location /path /path  {
+            client_max_body_size 1g;
+            index index.html default.html;
+        }
+	}
+}
```
gwolf-011235, line 12: P3: missing newline

src/ConfigFileParser.cpp
```diff
@@ -237,7 +240,9 @@ void ConfigFileParser::readServerBlock(void)
 	size_t startIndex = m_configFileIndex;
 
 	while (m_configFileContent[m_configFileIndex] != '}') {
-		if (isKeyword(convertBlockToString(LocationBlock), m_configFileIndex) && isValidBlockBeginn(LocationBlock)) {
+		if (isKeyword(convertBlockToString(LocationBlock), m_configFileIndex)) {
+			if (!isValidBlockBeginn(LocationBlock))
```
gwolf-011235, line 244: P3: isValidBlockBeginn --> should be "Begin"

gwolf-011235 (approved):
LGTM
Just some P3

config_files/location_empty_path.conf
QCHR1581: Done

config_files/location_multiple_paths.conf
QCHR1581: Done

src/ConfigFileParser.cpp
QCHR1581: Done

PR #157, opened 2025-01-25
fix: don´t try to send timeout messages to clients while CGI is hanging

### LINKED ISSUE #153: Timeout for Connection waiting on CGI
checkForTimeout() checks timeSinceLastEvent of a Connection. If it's more than timeout value Connection Status is set to Timeout, and fd is modified to EPOLLOUT (to send timeout message).

This is independent of Status. If a Connection is waiting for a CGI process to finish, it can also be set to Timeout. But it's fd is currently not monitored by epoll, so modifying it doesn't work.

We could exclude Status::SendToCGI and Status::ReceiveFromCGI. However then we don't have a way to remove a "hanging" CGI process.

Maybe other problem: the timeout message is only sent if the timed out fd triggers epoll again. A trigger does not mean that the fd is ready to receive something, but that it will not block > e.g. [LINK](https://stackoverflow.com/a/52976327) at the end.

I'm not sure if a fd could be in a state where it wont trigger epoll but still "hangs". I'm also not sure how we could test this.

u413-284-si:
Add check to not try to modify fd when CGI is the reason for timeout

closes #153

### DIFF
--- a/src/Server.cpp
+++ b/src/Server.cpp
@@ -1044,7 +1044,7 @@ void handleCompleteRequestHeader(Server& server, int clientFd, Connection& conne
 	if (connection.m_request.method == MethodPost && connection.m_request.isDirectory)
 		connection.m_request.httpStatus = StatusOK;
 
-    // FIXME:
+	// FIXME:
 	// Allow directories to be deleted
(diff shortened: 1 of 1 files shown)


gwolf-011235 (approved):
LGTM

PR #159, opened 2025-01-25
142 setup testing with siege

gwolf-011235:
Adds new Makefile target "test3" to run stresstest with siege.
Adds " test4" to benchmark the empty testfile

- new subdirectory /test/siege
- default siege.conf with one change: internet: true (siege accesses url at random)
-  urls.txt with many of our example html pages. Didn't include: 
   - delete.html: can't send delete request 
   - images_external.html: images are served by other server
   - run_cgi.html: cgi scripts are accessed directly
   - submit.html: can't send form data
   - upload.html: can't send form data
- Makefile sets some variables for testrun with Siege

Also split external images from images.html into new images_external.html. Updated styles.css to reduce duplication
Also added empty.html as testfile


u413-284-si (changes requested):
nice nice, would add this eval case:

```
Availability should be above 99.5% for a simple GET on an empty page with a siege -b on that
page.
```

QCHR1581 (approved):
LGTM!

gwolf-011235:
Added empty.html and test4 to benchmark the empty testfile

u413-284-si (approved):

PR #160, opened 2025-01-25
155 multiple root directives within one server/location should be invalid

### LINKED ISSUE #155: Multiple root directives within one server/location should be invalid
The configs below cause currently NOT an error message:

### Server contains multiple root directives

```nginx
server {
           root /var/www/html;
           root /var/www/project_red;
}
```
### Location contains multiple root directives

```nginx
server {
           location /error {
               root /var/www/html;
               root /var/www/error;
            }
}
```
Nginx does not accept both config files.
Therefore our webserver should also display an error message for such cases.

QCHR1581:
Refer to issue #155 for more information.

# Additional vars

Following variables were added:

- ``m_serverRootCount``
- ``m_locationRootCount``

Their purpose is to keep track of the count of roots within a server block and a location block. 

# Usage of additional vars

Before processing a server block the variable ``m_serverRootCount`` is set to zero. 
This is the same case for the variable ``m_locationRootCount`` besides before a location block. 

### processServerContent:
```cpp
	m_serverRootCount = 0;
	while (readAndTrimLine(serverBlockConfig.serverBlockContent, ';'))
		readServerConfigLine();

	m_isDefaultLocationDefined = false;
	for (std::vector<std::string>::const_iterator it = serverBlockConfig.locationBlocksContent.begin();
		 it != serverBlockConfig.locationBlocksContent.end(); ++it) {
		m_locationRootCount = 0;
		processLocationContent(*it);
	}
```

Before entering the function for reading the path of the root directive, the values of ``m_serverRootCount`` and ``m_locationRootCount`` get incremented.

### readServerDirectiveValue:

```cpp
if (directive == "listen")
	readListen(value);
else if (directive == "root") {
	m_serverRootCount++;
	readRootPath(ServerBlock, value);
}
```

### readLocationDirectiveValue:

```cpp
if (directive == "root") {
	m_locationRootCount++;
	readRootPath(LocationBlock, value);
} else if (directive == "alias")
	readAliasPath(value);
```

# Additional checks

Before checking if the path of the root directive is correct and then read, the values of ``m_serverRootCount`` and ``m_locationRootCount`` get checked.

### readRootPath:

```cpp
if (m_serverRootCount > 1)
	throw std::runtime_error(ERR_SERVER_MULTIPLE_ROOTS);
if (m_locationRootCount > 1)
	throw std::runtime_error(ERR_LOCATION_MULTIPLE_ROOTS);
```

Close #155

### DIFF
--- a/inc/ConfigFileParser.hpp
+++ b/inc/ConfigFileParser.hpp
@@ -44,6 +44,8 @@ private:
 	size_t m_serverIndex;
 	std::vector<std::string> m_validServerDirectives;
 	size_t m_locationIndex;
+	int m_serverRootCount;
+	int m_locationRootCount;
 	std::vector<std::string> m_validLocationDirectives;
 	bool m_isDefaultLocationDefined;
 	static const char* const s_whitespace;
--- a/inc/ConfigFileParser.hpp
+++ b/inc/ConfigFileParser.hpp
@@ -44,8 +44,8 @@ private:
 	size_t m_serverIndex;
 	std::vector<std::string> m_validServerDirectives;
 	size_t m_locationIndex;
-	size_t m_hasServerRoot;
-	size_t m_hasLocationRoot;
+	bool m_hasServerRoot;
+	bool m_hasLocationRoot;
 	std::vector<std::string> m_validLocationDirectives;
 	bool m_isDefaultLocationDefined;
 	static const char* const s_whitespace;
--- a/inc/ConfigFileParser.hpp
+++ b/inc/ConfigFileParser.hpp
@@ -44,8 +44,8 @@ private:
 	size_t m_serverIndex;
 	std::vector<std::string> m_validServerDirectives;
 	size_t m_locationIndex;
-	size_t m_serverRootCount;
-	size_t m_locationRootCount;
+	size_t m_hasServerRoot;
+	size_t m_hasLocationRoot;
 	std::vector<std::string> m_validLocationDirectives;
 	bool m_isDefaultLocationDefined;
 	static const char* const s_whitespace;
--- a/inc/ConfigFileParser.hpp
+++ b/inc/ConfigFileParser.hpp
@@ -44,8 +44,8 @@ private:
 	size_t m_serverIndex;
 	std::vector<std::string> m_validServerDirectives;
 	size_t m_locationIndex;
-	int m_serverRootCount;
-	int m_locationRootCount;
+	size_t m_serverRootCount;
+	size_t m_locationRootCount;
 	std::vector<std::string> m_validLocationDirectives;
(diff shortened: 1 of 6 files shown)


inc/ConfigFileParser.hpp
```diff
@@ -44,6 +44,8 @@ class ConfigFileParser {
 	size_t m_serverIndex;
 	std::vector<std::string> m_validServerDirectives;
 	size_t m_locationIndex;
+	size_t m_serverRootCount;
+	size_t m_locationRootCount;
```
gwolf-011235, line 48: P3: Since we never expect more than one I think a bool m_hasServerRoot and bool has_LocationRoot would make more sense

gwolf-011235 (approved):
LGTM

A P3 suggestion

inc/ConfigFileParser.hpp
QCHR1581: True, those names fit better. Done

QCHR1581: True, those names fit better. Done

gwolf-011235: I meant to change type to `bool` :D

u413-284-si (approved):
P2: You could also just use booleans to indicate whether a root or location directive has already been found. We don't care how many, just 0 or 1. If 1 and another comes along we throw.

It is a suggestion though, no need to change

inc/ConfigFileParser.hpp
QCHR1581: Ohhhhh, makes sense :sweat_smile:.

Adopted also ``readRootPath`` accordingly:
```cpp
void ConfigFileParser::readRootPath(const Block& block, std::string rootPath)
{
	if (block == ServerBlock && m_hasServerRoot)
		throw std::runtime_error(ERR_SERVER_MULTIPLE_ROOTS);
	if (block == LocationBlock && m_hasLocationRoot)
		throw std::runtime_error(ERR_LOCATION_MULTIPLE_ROOTS);

	if (rootPath.find_first_of(s_whitespace) != std::string::npos)
		throw std::runtime_error(ERR_MULTIPLE_ROOT_PATHS);

	if (rootPath.at(0) != '/')
		throw std::runtime_error(ERR_ROOT_PATH_MISSING_SLASH);

	if (rootPath[rootPath.length() - 1] == '/')
		rootPath.erase(rootPath.end() - 1);

	if (block == ServerBlock) {
		m_hasServerRoot = true;
		m_configFile.servers[m_serverIndex].root = rootPath;
	} else if (block == LocationBlock) {
		m_hasLocationRoot = true;
		m_configFile.servers[m_serverIndex].locations[m_locationIndex].root = rootPath;
	}
}
```

PR #161, opened 2025-01-25
156 move config files used for testing to test folder and create two config files for showcasing

### LINKED ISSUE #156: Move config files used for testing to test folder and create two config files for showcasing
## Move config files used for testing

Currently config files which are used for testing purposes are stored in the folder config_files.

Those files should be moved to a folder with the same name within the test folder. 

## Config files for showcasing

In the config_files folder should only be two config files:

### 1. trihard.conf: 
Should be used when starting the webserv without any config as default config. 
Contains only minimum setup comparable to nginx's default config file. 

### 2. example.conf:

Replaces the standard_config.conf file for showcasing what the server is capable of.
Should also be used for our integration tests.

QCHR1581:
- Moved config files which are used for testing purposes to a folder named **config_files** within the **test** directory.
- Created the config file **trihard.conf** which is used as default config file for webserv
- Renamed **standard_config.conf** to **example.conf** which gets used in integration tests and for showcasing the capability of webserv

Close #156


gwolf-011235 (changes requested):
I would delete target "run" as we can just ./webserv now
Then we only have one CONFIGFILE variables aka delete CONFIGFILE_INTEGRATION
You need to update example.conf with values from integration_test_config.conf (I changed it and added some stuff)

QCHR1581:
> I would delete target "run" as we can just ./webserv now Then we only have one CONFIGFILE variables aka delete CONFIGFILE_INTEGRATION You need to update example.conf with values from integration_test_config.conf (I changed it and added some stuff)

The target run from the makefile is deleted as well as CONFIGFILE_INTEGRATION (CONFIGFILE gets used instead). 

I already updated the example.conf file with the values from the integration_test_config.conf

gwolf-011235 (approved):
LGTM

PR #163, opened 2025-01-25
162 kiil timed out cgi processes

### LINKED ISSUE #162: Kiil timed out CGI processes
- Create a destructor of Connection which
  - kill() CGI process if pid is not -1
  - closes CGI pipes
  - closes client fd

- removeConnection() doesn't need to close fd since it should be done in destructor

- clearConnection() doesn't need to close CGI fds since it is only called if everything went well and connection header is keep-alive

- reset m_pid in connectionSendToCGI() --> if we harvest process we can reset it

- checkForTimeout() should 
  - if not CGI: modifyEvent 
  - if CGI: addEvent clientFD and removeEvent CGI pipes

u413-284-si:
Implements points listed in #162 

closes #162

### DIFF
--- a/inc/CGIHandler.hpp
+++ b/inc/CGIHandler.hpp
@@ -59,7 +59,7 @@ private:
 	std::string extractPathInfo(const std::string& path);
 	std::string extractScriptPath(const std::string& path);
 	void closePipes();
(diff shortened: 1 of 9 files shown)


gwolf-011235 (approved):
LGTM

PR #169, opened 2025-01-26
164 move test header files into test directory

gwolf-011235:
1. Restructure
- Moves all unit test source files in subdir test/unit/src
- Moves all unit test header files in subdir test/unit/inc
- Adapt Makefile to find new location of src and header

2. Adapt `test` target
- name variable of unittest is now UNIT instead of TEST
- make test runs UNIT (instead of just compiling it). It has ingredient UNIT so it will also compile if it does not exist
- adapted Makefile for name change


u413-284-si (changes requested):
As such looks good, but siege is not working anymore. The problem exists also in main. Branch 141 is still functioning.

gwolf-011235:
Problem is that test3 still uses deprecated CONFIGFILE_INTEGRATION --> will fix in here

gwolf-011235:
Changed CONFIGFILE_INTEGRATION to CONFIGFILE --> test3 and test4 can be used again

u413-284-si (approved):

PR #170, opened 2025-01-26
167 adapt status code functions

gwolf-011235:
- extract conversion logic out of ostream inserter into own function statusCodeToString. Use this function in ostream inserter
- added return for both statusCodeToString and statusCodeToReasonPhrase in case of no switch case matches
- refactor stringToStatusCode to convert the string and the check with switch case
   - with str.size() != 3 we can make fast check if it can be valid status code, as every status code is 3 numbers long
   - with switch case we get compiler warning if we add a new status code --> we can't forget 
- update Doxgen

### DIFF
--- a/inc/StatusCode.hpp
+++ b/inc/StatusCode.hpp
@@ -21,10 +21,11 @@ enum statusCode {
 	StatusNonSupportedVersion = 505
 };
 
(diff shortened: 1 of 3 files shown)


u413-284-si (approved):
Is ok, just needs the siege fix

PR #171, opened 2025-01-26
166 set shallcloseconnection in responsebuilder

gwolf-011235:
- new function isCloseConnectionStatus() which checks if request has closing status and sets bool
- implemented in ResponseBuilder as first check
- removed `request.shallCloseConnection = true` lines in RequestParser and Server 
- adapted integration tests to check for header connection


u413-284-si (approved):
LGTM

PR #174, opened 2025-01-26
173 add valgrind test with automated request delivery

### LINKED ISSUE #173: Add valgrind test with automated request delivery
We want to be able to check for leaks/invalid writes while having the webserver be targeted with multiple various requests

u413-284-si:
Add python script to send multiple requests in a random order (GET/POST) for a time frame of 30s. One DELETE request gets send afterwards. 

Add text file to dictate the requests to be sent.

Add target "test5" to Makefile.

Fix bug: After accepting Content-Length 0, the server also expected a body. In case of a DELETE request no body is send, resulting in the server waiting for a body but not receiving any. Add check to only expect body if Content-Length != 0.

closes #173


Makefile
```diff
@@ -316,6 +317,21 @@ test4: $(NAME) | $(LOG_DIR)
 		$(SIEGE_URL)
 	$(SILENT)kill `cat webserv.pid` && rm -f webserv.pid
 
+REQUESTER=$(REQUESTER_DIR)/requester.py
+REQUESTER_FILE=$(REQUESTER_DIR)/requests.txt
```
gwolf-011235, line 321: Since we have .txt in gitignore this wasn't uploaded. 
You need to delete from gitignore, stage the .txt and then add back to gitignore.
You don't even need to make a commit for the gitignore change, just delete it to be able to stage the .txt

Makefile
```diff
@@ -316,6 +317,21 @@ test4: $(NAME) | $(LOG_DIR)
...
+
+.PHONY: test5
+test5: $(NAME) | $(LOG_DIR)
+	@printf "$(YELLOW)$(BOLD)Run webserv with valgrind and requester$(RESET) [$(BLUE)$@$(RESET)]\n"
+	$(SILENT)valgrind ./webserv $(CONFIGFILE) >$(LOG_VALGRIND) 2>&1 & echo $$! > webserv.pid
```
gwolf-011235, line 326: Would add valgrind options from deleted target valgr like --leak-check=full. Also add --track-fds=yes (which we didn't have in target)
Could store them in a variable VALGRIND_OPTIONS or something

Makefile
```diff
@@ -316,6 +317,21 @@ test4: $(NAME) | $(LOG_DIR)
...
+	$(SILENT)sleep 1
+	$(SILENT)/usr/bin/python3 \
+		$(REQUESTER) \
+		$(REQUESTER_FILE)
+	$(SILENT)kill `cat webserv.pid` && rm -f webserv.pid
+	$(SILENT)sleep 2
```
gwolf-011235, line 332: Why sleep here again?

gwolf-011235 (changes requested):
.txt file is missing.

I haven't looked at the python script, I trust :D

Makefile
u413-284-si: Weird, I did exactly this, somehow didnt work the first time. Now it's there

Makefile
u413-284-si: So you can see the result of the python script

Makefile
u413-284-si: Added flags

gwolf-011235: Missing:
--track-fds=yes [to track fd]
--log-file=$(LOG_VALGRIND) [to write valgrind output to logfile]

Then we can redirect output of webserv to its own logfile LOG_WEBSERV:
`$(SILENT)valgrind $(VALGRIND_FLAGS) ./webserv $(CONFIGFILE) > $(LOG_WEBSERV) 2>&1 & echo $$! > webserv.pid`

Output of valgrind is in valgrind log
Output of webserv is in webserv log (right now /dev/null)

u413-284-si: Added mentioned flags

gwolf-011235 (approved):
LGTM

PR #175, opened 2025-01-26
172 implement static values for epoll

gwolf-011235:
- Implemented s_maxEvents and s_epollTimeout and use them in main
- Also moved init of logger up to be able to log configfile path.
- Little bugfix: while checking for error in checkForTimeout (missing brackets) I moved setting of status Timeout too high. We never entered the CGI cases --> fixed
- add new integration test_timeout.py. Tries to simulate timeout. Since we don't have way to dynamically change timeout value these take very long (60 sec a pop). Added marker and deactivated timeout tests for "normal"  test run --> -m "not timeout" in pytest.ini. If removed will also run timeout tests
- added script infinite.sh, which just has infinite loop --> used for testing

### DIFF
--- a/cgi-bin/infinite.sh
+++ b/cgi-bin/infinite.sh
@@ -1,0 +1,5 @@
+#!/bin/bash
+
+while true; do
+    sleep 1
+done
--- a/inc/Server.hpp
+++ b/inc/Server.hpp
@@ -50,6 +50,8 @@
  */
 class Server {
 public:
+	static const int s_maxEvents = 10; /**< Maximum number of events to process via epoll */
+	static const int s_epollTimeout = 1000; /**< Timeout for epoll_wait in milliseconds */
 	static const int s_backlog = 10; /**< Default backlog for listening sockets */
 	static const time_t s_clientTimeout = 60; /**< Default timeout for a Connection in seconds */
 	static const std::size_t s_bufferSize = 1024; /**< Default buffer size for reading from sockets in Bytes */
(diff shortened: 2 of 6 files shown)


u413-284-si (changes requested):
The infinite loop timeout test fails with a bad fd :(

gwolf-011235:
I think this was just flaky test: we have timeout set at 60 seconds and I had test timeout at 61 seconds - so not much room for error. I set timeout in test to 65 (which means test take longer but shouldn't fail)

I'm pretty sure bad fd came from epoll trying to remove CGI file descriptor, which is still registered. Which didn't make test fail, but was just the last thing server reported.
I changed webutils::closeFd to removeCGIFiledescriptor() --> I think we can remove them at timeout stage (if they exist)
I also adapted epoll remove event to only send WARN message, so it doesn't look so scary.
And I init time string buffer to zero - not sure if it makes a difference but shouldn't hurt either.

u413-284-si:
Looks good now. Just there should be a better way to activate the timeout tests other than deleting the marker from pytest.ini, no?

u413-284-si (approved):

gwolf-011235:
- removed addopts from pytest.ini
- added automatic marker generation based on test name to conftest.py
- added markers to pytest.ini
- added variable PYTEST to test2 which contains addopts with default value `-v -m "not timeout"`

We can change on cli: `make test2 PYTEST=`  to run all or `make test2 PYTEST="-m cgi"` to run all cgi tests (n.b.: needs quoting)

PR #176, opened 2025-01-26
165 trimtrailingwhitespace should return string

gwolf-011235:
- trimtrailingwhitespace now returns std::string. It can be chained.
- added convenience function trimWhiteSpaces which trims leading and trailing
- implemented trimWhiteSpaces
- removed timeStringBuffer - was a global :o Made own value in respective functions


u413-284-si (approved):
LGTM

PR #177, opened 2025-01-26
168 requestparserparseheader should not set status

### LINKED ISSUE #168: RequestParser.parseHeader should not set status
Right now in case of error we 
- set status to StatusBadRequest
- set shallCloseConnection to true
- throw with error message

We could also just throw and set the status in catch block of the calling server function.

If we refactor away shallCloseConnection ( see #166 ) many function signatures can be adapted to not receive HTTPRequest anymore.

u413-284-si:
Move setting of httpStatus into catch blocks
Add custom exceptions to handle statii other than BadRequest
Adapt unit tests

closes #168

### DIFF
--- a/src/Server.cpp
+++ b/src/Server.cpp
@@ -1012,8 +1012,33 @@ void handleCompleteRequestHeader(Server& server, int clientFd, Connection& conne
 
 	try {
 		server.parseHeader(connection.m_buffer, connection.m_request);
+	} catch (RequestParser::MethodNotImplementedException& e) {
+		LOG_ERROR << e.what();
+		connection.m_request.httpStatus = StatusMethodNotImplemented;
+		connection.m_status = Connection::BuildResponse;
+		server.modifyEvent(clientFd, EPOLLOUT);
+		return;
+	} catch (RequestParser::MethodNotAllowedException& e) {
+		LOG_ERROR << e.what();
+		connection.m_request.httpStatus = StatusMethodNotAllowed;
+		connection.m_status = Connection::BuildResponse;
+		server.modifyEvent(clientFd, EPOLLOUT);
+		return;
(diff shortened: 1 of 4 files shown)


src/Server.cpp
```diff
@@ -1012,8 +1012,33 @@ void handleCompleteRequestHeader(Server& server, int clientFd, Connection& conne
 
 	try {
 		server.parseHeader(connection.m_buffer, connection.m_request);
+	} catch (RequestParser::MethodNotImplementedException& e) {
+		LOG_ERROR << e.what();
+		connection.m_request.httpStatus = StatusMethodNotImplemented;
+		connection.m_status = Connection::BuildResponse;
+		server.modifyEvent(clientFd, EPOLLOUT);
```
gwolf-011235, line 1019: P3: We use server.modifyEvent() 22 times. Nearly everytime we 
- change connection status
- modify client fd 
We could make a little refactor and introduce function changeConnectionState()

gwolf-011235 (approved):
Looks good, just some P3s

gwolf-011235:
New Idea: 
Since only difference for custom exception type is status code we could customize an exception to hold a statusCode. Then we can throw one custom exception and just extract statusCode:
```
struct HTTPErrorException : public std::runtime_error {
  explicit HTTPErrorException(statusCode statusCode, const std::string& msg);
  statusCode statusCode;
};
```
Written: `throw HTTPErrorException (StatusMethodNotImplemented, ERR_METHOD_NOT_IMPLEMENTED);`

And catch block:
```
catch (const RequestParser::HTTPErrorException & e) {
  LOG_ERROR << e.what();
  connection.m_request.httpStatus = e.statusCode;
  connection.m_status = Connection::BuildResponse;
  server.modifyEvent(clientFd, EPOLLOUT);
  return;
} 
```

Would reduce number of catch blocks

PR #180, opened 2025-01-27
179 fix directive inheritance

### LINKED ISSUE #179: fix directive inheritance
# Current implementation

Currently a location can inherit the values of the following directives defined in a server block:

- root
- client_max_body_size
- error_page

It is implemented in this way:

```cpp
if (location.root == "html")
	location.root = server.root;
else if (location.maxBodySize == constants::g_oneMegabyte)
	location.maxBodySize = server.maxBodySize;
else if (location.errorPage.empty())
	location.errorPage = server.errorPage;
```

# Problem

The inheritance of those directives does not work correctly, when a value for one of the mentioned directives is provided, which is the same as the default value of the location directive. 

Example:

```nginx
server {
              client_max_body_size 2m;
              location {
                               client_max_bodysize 1m;
                             } 
}
```

In this case the directive ``client_max_body_size`` within the location block would get the same value as this directive within the server block. 
This is caused by the fact, that the value ``1m`` in the location block is also the default value of the directive ``client_max_body_size``.

QCHR1581:
Refer to issue #179 for more information.

# Additional var

I added the var ``m_hasLocationMaxBodySize`` to indicate if a value for the directive client_max_body_size was specified within the config file.

# Refactor

When now a value is succesfully read for the directives:

- root
- client_max_body_size

within a location, the vars ``m_hasLocationRoot`` and ``m_hasLocationMaxBodySize`` are set to true. 

Therefore the checks, if those directives should inherit the values from the server block, were adapted:

```cpp
if (!m_hasLocationRoot)
	location.root = server.root;
if (!m_hasLocationMaxBodySize)
	location.maxBodySize = server.maxBodySize;
if (location.errorPage.empty())
	location.errorPage = server.errorPage;
```

# Test

The test for the directive inheritance was updated by including another location. 
This additional location specifies the value ``1m`` for the directive client_max_body_size. 
The expected behaviour is that this directive keeps the value ``1m`` and does not inherit the value from the server.

Closes #179

### DIFF
--- a/src/ConfigFileParser.cpp
+++ b/src/ConfigFileParser.cpp
@@ -347,7 +347,7 @@ void ConfigFileParser::processLocationContent(const std::string& locationBlockCo
 	ConfigServer& server = m_configFile.servers[m_serverIndex];
 	Location& location = server.locations[m_locationIndex];
 
-	if (location.root != "html" && !location.alias.empty())
+	if (m_hasLocationRoot && !location.alias.empty())
 		throw std::runtime_error(ERR_ROOT_AND_ALIAS_DEFINED);
 	if (!m_hasLocationRoot)
 		location.root = server.root;
--- a/src/ConfigFileParser.cpp
+++ b/src/ConfigFileParser.cpp
@@ -14,6 +14,7 @@ ConfigFileParser::ConfigFileParser(void)
 	, m_locationIndex(0)
 	, m_hasServerRoot(false)
 	, m_hasLocationRoot(false)
+	, m_hasLocationMaxBodySize(false)
 	, m_isDefaultLocationDefined(false)
 {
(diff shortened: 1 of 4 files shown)


src/ConfigFileParser.cpp
```diff
@@ -347,11 +349,11 @@ void ConfigFileParser::processLocationContent(const std::string& locationBlockCo
 
 	if (location.root != "html" && !location.alias.empty())
```
gwolf-011235, line 350: Couldn't the check `if (location.root != "html" && !location.alias.empty())` also use m_hasLocationRoot for the case that root defines the default value and an alias directive is defined?

gwolf-011235 (approved):
Looks good, just one question

PR #181, opened 2025-01-27
178 add integration tests

gwolf-011235:
Added new integration tests
fixed two bugs:
- dont always change target resource in delete / file handler
- uncomment break in delete handler


u413-284-si (approved):
LGTM

